PR #54, opened 2020-08-31
Update app.start()'s message and docstring

seratch:
`App#start()` uses `http.server` module for starting a web server. Using the module in production is [not recommended](https://docs.python.org/3/library/http.server.html).

> http.server is not recommended for production. It only implements basic security checks.

This pull request makes `App#start()`'s message and related comments in `App` clearer.

I think depending on a specific WSGI server just for offering a production-ready `App#start()` method is not a great idea. Some may prefer Gunicorn while others may want to use uWSGI. This product should not narrow any developers' options. Thus, implying the necessity to use a production WGSI server by having `(development server)` in the message and renaming the server implementation to `SlackAppDevelopmentServer` should be a reasonable option.

`AsyncApp#start()` starts AIOHTTP's web server. Using this in production doesn't have any issues. So, I don't apply the same changes to `AsyncApp`.

* [x] I've run `./scripts/run_tests.sh` after making the changes.

### DIFF
--- a/slack_bolt/app/app.py
+++ b/slack_bolt/app/app.py
@@ -260,10 +260,10 @@ class App:
     # standalone server
 
     def start(self, port: int = 3000, path: str = "/slack/events") -> None:
-        self.server = SlackAppServer(
+        self._development_server = SlackAppDevelopmentServer(
             port=port, path=path, app=self, oauth_flow=self.oauth_flow,
         )
-        self.server.start()
+        self._development_server.start()
 
     # -------------------------
     # main dispatcher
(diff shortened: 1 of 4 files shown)


slack_bolt/app/app.py
```diff
@@ -799,10 +799,20 @@ def _register_listener(
...
+        """Slack App Development Server
+
+        This is a thin wrapper of http.server.HTTPServer and is good enough
+        for your local development or prototyping.
+
+        However, as mentioned in Python official documents, using http.server module in production
+        is not recommended. Please consider using an adapter (refer to slack_bolt.adapter.*)
+        along with a production-grade server when running the app for end users.
+        https://docs.python.org/3/library/http.server.html#http.server.HTTPServer
+        """
```
mwbrooks, line 815: I really like the detail that's explained here rather than writing it as documentation that could be easily overlooked.

Speaking as someone less experienced with Python, does redirecting the developer to `slack_bolt.adapter.*` give them enough information to add support for a production server? Or do we need additional documentation to guide them through adding an adapter for a production-grade server?

seratch: I think we should have a page (or a section) about the production server in the documents. The section doesn't exist in Bolt for JS documents as it supports only Express.js and it's totally fine to use its server implementation (similarly to AIOHTTP, Sanic in Python). Once we come up with the page, I would like to have a link to the document here.

Issue #98, opened 2020-09-22
Add authorize function support


seratch:
The [`authorize` function that is available in Bolt JS](https://slack.dev/bolt-js/concepts#authorization) is not yet supported in Bolt for Python. 

```python
# these arguments should be injected in correspondence with their names
# also, the Bolt framework provides `args` as a single value as with middleware/listeners
def authorize(enterprise_id, team_id, user_id):
    return {
        "bot_id": "B111",
        "bot_user_id": "W111",
        "bot_token": "xoxb-"
    }

app = App(
    signing_secret: os.environ["SLACK_SIGNING_SECRET"],
    authorize: authorize
)
```

For most use cases, probably, developers prefer using `installation_store`. That being said if a developer would like to implement a simple but more flexible way to fetch installation data, going with the `authorize` function can be a handy way.

### Category (place an `x` in each of the `[ ]`)

* [x] **slack_bolt.App** and/or its core components
* [x] **slack_bolt.async_app.AsyncApp** and/or its core components
* [ ] Adapters in **slack_bolt.adapter**
* [ ] Others

## Requirements

Please read the [Contributing guidelines](https://github.com/slackapi/bolt-python/blob/main/.github/contributing.md) and [Code of Conduct](https://slackhq.github.io/code-of-conduct) before creating this issue or pull request. By submitting, you are agreeing to those rules.

PR #99, opened 2020-09-22
Make single team auth Bolt JS compatible

seratch:
This pull request modifies `SingleTeamAuthorization` middleware to be compatible with Bolt for JS. Now the `App` calls `auth.test` API method when booting a Bolt app and raises an exception if the call fails. 

Also, this pull request removes the `authorization_test_enabled` option and we will be rethinking a better way to provide the option for multiple workspace apps. Probably, one of the options developers can take advantage of would be to utilize `authorize` function described at #98 .

* [ ] Others

### DIFF
--- a/slack_bolt/app/async_app.py
+++ b/slack_bolt/app/async_app.py
@@ -85,7 +85,6 @@ class AsyncApp:
         # for the OAuth flow
         oauth_settings: Optional[AsyncOAuthSettings] = None,
         oauth_flow: Optional[AsyncOAuthFlow] = None,
-        authorization_test_enabled: bool = True,
         # No need to set (the value is used only in response to ssl_check requests)
         verification_token: Optional[str] = None,
     ):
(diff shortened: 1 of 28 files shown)


slack_bolt/app/async_app.py
```diff
@@ -188,18 +183,13 @@ def _init_async_middleware_list(self):
         )
         if self._async_oauth_flow is None:
             if self._token:
-                self._async_middleware_list.append(
-                    AsyncSingleTeamAuthorization(
-                        verification_enabled=self._authorization_test_enabled
-                    )
-                )
+                self._async_middleware_list.append(AsyncSingleTeamAuthorization())
```
seratch, line 186: To avoid depending on a specific event loop here, we don't do a blocking operation here (NOTE: `__init__` is not an async function. We cannot have `await` here).

tests/scenario_tests/test_attachment_actions.py
```diff
@@ -105,7 +105,7 @@ def test_failure_without_type(self):
         app.action("unknown")(simple_listener)
         response = app.dispatch(request)
         assert response.status == 404
-        assert self.mock_received_requests["/auth.test"] == 2
+        assert self.mock_received_requests["/auth.test"] == 1
```
seratch, line 108: Now that the `auth.test` API call result is cached, the call count must be always 1 for the same token.

seratch (commented):
Comments for reviewers

seratch:
To: reviewers
Let me merge this change now. If you have comments / feedback, please let me know by writing comments at any time. I will quickly fix them by the first beta release.

PR #126, opened 2020-10-25
Add app_id in AuthorizeResult (InstallationStoreAuthorize)

seratch:
The way to access `app_id` is tricky in some cases. In many features (e.g., Events API, Interactivity payloads), the field is available as `api_app_id` at the top level of request body. However, in the case of message actions, the field is available in `message.bot_profile.app_id`.

This pull request adds `app_id` in `AuthorizeResult` when Bolt framework can get the value without extra cost. Specifically, developers can expect the field is available in `AuthorizeResult` when using `InstallationStoreAuthorize` after applying this change. Of course, there is no blocker if a developer would like to set `app_id` in `AuthorizeResult` in their own custom `authorize` functions.

Considering more easiness to access the value in requests, we may want to expose the value in requests as a top-level field in `context` objects. While we can do that as well, it's not in the scope of this pull request so far. If others think we should do that too, I can work on it. Personally, I'm happy to add it too.

* [ ] Others


seratch:
For context, I wanted to easily access `app_id` when building a deep link in a message - `slack://app?team={team_id}&id={app_id}`. In the case, checking the way to access `app_id` in payloads was a bit confusing. I think this is a common use case.

stevengill (approved):

stevengill:
To keep consistency, is this something we should add to AuthorizeResult in bolt-js and bolt-java?

stevengill:
When we do oauth in JS land, we do have `app_id` part of the Installation object. But we don't return it AuthorizeResult when calling the built in oauth authorize function

aoberoi:
@seratch thanks for motivating this with a use case, that makes it much easier to understand. if there are any other use cases you have in mind, that might also help.

my main question is: why is producing an `app_id` the responsibility of `authorize`?

> Considering more easiness to access the value in requests, we may want to expose the value in requests as a top-level field in context objects. 

IMHO, this is a more "idiomatic" approach. in Bolt for JS, we use default global middleware to process incoming payloads and make data available in a consistent place within `context`. this type of operation seems to fit that pattern well.

The `authorize` function is about producing data that is not already in the payload, for which a store can be queried, and for which values are varied per-installation. The `app_id` is static within an `App` instance, so I don't think it makes sense to ask `authorize` to produce it.

if we don't want to use a default global middleware, you could also use a constant in the app. this might be reasonable if there aren't too many other use cases besides producing a deep link that we can think of.

seratch:
@stevengill Thanks for the comment. Considering Ankur's suggestion, we can revisit the details of the implementation once we decide implementing this.

@aoberoi I think it's also great if we can agree we should have `app_id` in `context`. If we should choose the only a single best answer, I don't have any strong reason to choose authorize over context here.

Having the value as a static data requires an additional `bots.info` call or asking developers to set the value when booting the app. So, I prefer extracting app_id from request body and set it in context for this.

If we can agree to add the value to context, I'm happy to close this PR and will come up with another one to add the value in context.

seratch:
I've created an issue for implementing `app_id` in `context`: https://github.com/slackapi/bolt-python/issues/129

Let me close this PR now.

seratch:
I would like to revisit this - for context: https://github.com/slackapi/bolt-python/issues/129#issuecomment-717174392

seratch:
Closing this PR - refer to https://github.com/slackapi/bolt-python/issues/129#issuecomment-753801091 for its context

PR #148, opened 2020-11-13
Add org-level installation support

stevengill:
(Describe the goal of this PR. Mention any related Issue numbers)
https://github.com/slackapi/python-slack-sdk/pull/877
### Category (place an `x` in each of the `[ ]`)

* [ ] `slack_bolt.App` and/or its core components
* [ ] `slack_bolt.async_app.AsyncApp` and/or its core components
* [ ] Adapters in `slack_bolt.adapter`
* [ ] Document pages under `/docs`
* [x] Others

## Requirements (place an `x` in each `[ ]`)

Please read the [Contributing guidelines](https://github.com/slackapi/bolt-python/blob/main/.github/contributing.md) and [Code of Conduct](https://slackhq.github.io/code-of-conduct) before creating this issue or pull request. By submitting, you are agreeing to those rules.

* [x] I've read and understood the [Contributing Guidelines](https://github.com/slackapi/bolt-python/blob/main/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).
* [x] I've run `./scripts/install_all_and_run_tests.sh` after making the changes.

PR #171, opened 2020-12-05
Add v1.0 compatible mode to #148 Org-wide App support

seratch:
This pull request adds a new option to OAuth flow supported apps. The option is `installation_store_bot_only: bool` on `App`/`AsyncApp` and `OAuthSettings`/`AsyncOAuthSettings`. 

The default is `False`. If the value is `False`, the `InstallationStore` uses only `#find_installation` method for fetching installation data for incoming requests as we changed since v1.1. If it's `True`, `#find_bot` method is used as with v1.0. The fallback to `#find_bot` method enables existing apps to successfully access the installation data generated by v1.0.x apps. 

Why is this needed? If an app uses Amazon S3 backend `InstallationStore`, the already stored data is not compatible with v1.1's default `authorize` logic and we are unable to make it compatible in an efficient way. That's why we need to add this compatibility mode. I've observed this situation with the Amazon S3 InstallationStore but any custom InstallationStore can be affected in the same way. It depends on how InstallationStore access the stored data.

With the change in this PR, those developers can easily switch to v1.0 compatibility mode just by having `installation_store_bot_only=True`. Also, this should be valuable not only for the v1.0 backward compatibility but also for developers that want to manage only bot related installation data.

ref: #148 

* [ ] Others

### DIFF
--- a/scripts/run_tests.sh
+++ b/scripts/run_tests.sh
@@ -20,7 +20,7 @@ else
     black slack_bolt/ tests/ \
       && pytest \
       && pip install -e ".[adapter]" \
-      && pip install -U pytype \
+      && pip install -U pip setuptools wheel pytype \
       && pytype slack_bolt/
   else
     black slack_bolt/ tests/ && pytest
--- a/slack_bolt/adapter/aws_lambda/lambda_s3_oauth_flow.py
+++ b/slack_bolt/adapter/aws_lambda/lambda_s3_oauth_flow.py
@@ -62,7 +62,9 @@ class LambdaS3OAuthFlow(OAuthFlow):
         # the settings may already have pre-defined authorize.
         # In this case, the /slack/events endpoint doesn't work along with the OAuth flow.
         settings.authorize = InstallationStoreAuthorize(
-            logger=logger, installation_store=settings.installation_store
+            logger=logger,
+            installation_store=settings.installation_store,
+            bot_only=settings.installation_store_bot_only,
         )
 
         OAuthFlow.__init__(self, client=client, logger=logger, settings=settings)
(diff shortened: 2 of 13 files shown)


scripts/run_tests.sh
```diff
@@ -20,7 +20,7 @@ else
     black slack_bolt/ tests/ \
       && pytest \
       && pip install -e ".[adapter]" \
-      && pip install -U pytype \
+      && pip install -U pip setuptools wheel pytype \
```
seratch, line 23: Unrelated to this main change in this PR. Installing pytype often fails due to old versions of setuptools and wheel. Doing this prevents the issue from happening.

slack_bolt/adapter/aws_lambda/lambda_s3_oauth_flow.py
```diff
@@ -62,7 +62,9 @@ def __init__(
         # the settings may already have pre-defined authorize.
         # In this case, the /slack/events endpoint doesn't work along with the OAuth flow.
         settings.authorize = InstallationStoreAuthorize(
-            logger=logger, installation_store=settings.installation_store
+            logger=logger,
+            installation_store=settings.installation_store,
+            bot_only=settings.installation_store_bot_only,
```
seratch, line 67: In `InstallationStore` class, `installation_store_` is redundant. So, I chose just `bot_only` for it.

slack_bolt/authorization/async_authorize.py
```diff
@@ -91,19 +91,24 @@ async def __call__(
 
 
 class AsyncInstallationStoreAuthorize(AsyncAuthorize):
-    authorize_result_cache: Dict[str, AuthorizeResult] = {}
```
seratch, line 94: This was unintentional data sharing among instances.

slack_bolt/authorization/async_authorize.py
```diff
@@ -123,7 +128,7 @@ async def __call__(
         bot_token: Optional[str] = None
         user_token: Optional[str] = None
 
-        if self.find_installation_available:
+        if not self.bot_only and self.find_installation_available:
```
seratch, line 131: If bot_only is True, `find_installation` is never used.

PR #250, opened 2021-03-02
Add 'channel_join' and 'group_join' to "kept" subtypes

eddyg:
These are useful with the Events API since there are no explicit `channel_joined` or `group_joined` events like there are when using the RTM interface.

* [X] `slack_bolt.async_app.AsyncApp` and/or its core components
* [ ] Others

### DIFF
--- a/slack_bolt/middleware/ignoring_self_events/ignoring_self_events.py
+++ b/slack_bolt/middleware/ignoring_self_events/ignoring_self_events.py
@@ -33,18 +33,25 @@ class IgnoringSelfEvents(Middleware):
     # but the user ID might match our own app. Filter these out.
     # However, some events still must be fired, because they can make sense.
     events_that_should_be_kept = ["member_joined_channel", "member_left_channel"]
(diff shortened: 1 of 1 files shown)


eddyg:
It's not clear to me if this change makes sense or not, so I'll leave it up to you guys to decide.

For example, is there an advantage to not requiring more OAuth scopes to detect this information?

I'm also wondering if this would be better served by being "flipped" so you only get these `channel_join` and `group_join` subtype events _for just the app_, as an easy way to know when it gets invited to a new channel.

eddyg:
Since `channel_join` and `group_join` subtype events aren't even listed [in the docs](https://api.slack.com/events/message) as "working with" the Events API (even though they are being sent right now), I'm going to close this issue.

I think the "right" way to do what I want is to subscribe to `member_joined_channel` events. It's easy enough to check if the event is for the app (the only thing I'm interested in) by checking if `event.get("user", "") == context.bot_user_id`.

Apologies for the PR noise.

PR #290, opened 2021-04-12
Fix #273 Enable developers to customize the way to handle unmatched requests

### LINKED ISSUE #273: Enable developers to customize the way to handle unmatched requests
If an incoming request does not match with any listeners in a Bolt app, the Bolt app responds to the request with 404 response status. The following lines of code are relevant parts in `App` / `AsyncApp`:

* https://github.com/slackapi/bolt-python/blob/v1.4.4/slack_bolt/app/app.py#L399-L400
* https://github.com/slackapi/bolt-python/blob/v1.4.4/slack_bolt/app/async_app.py#L438-L439

There is no way to append custom logic for the patterns. A common use case can be to customize the logging in the scenario.  

To improve this, I'm thinking to change the behavior as below:

* Add a new Bolt specific exception `BoltUnhandledRequestError(BoltError)` [here](https://github.com/slackapi/bolt-python/blob/v1.4.4/slack_bolt/error/__init__.py) and enable `@app.error` listener to handle the pattern
* Change `App` / `AsyncApp` internals. If a request is unhandled, `App`/`AsncApp` raises `BoltUnhandledRequestError` and lets `@app.error` listener  handle it. If `@app.error` listener does not return any `BoltResponse` or set status to the `response` argument, `App`/`AsncApp` falls back to the current logic, which outputs simple log message and returns 404 status anyway.
* As with the current implementation, non-200 responses are ignored in Socket Mode

If there is no concern about this change, we'll apply the change in the next minor version (1.5.0).

* [x] **slack_bolt.async_app.AsyncApp** and/or its core components
* [ ] Others

## Requirements

seratch:
This pull request resolves #273 by introducing a new option to enable using `@app.error` handlers for unmatched requests. The default is set to False, which is fully backward compatible.

If the option is True, Bolt raises a `BoltUnhandledRequestError` with sufficient information. `@app.error` handler can customize the behavior for the patterns (e.g., having custom logging, changing HTTP status from 404 to something else).

```python
app = App(
    token=os.environ["SLACK_BOT_TOKEN"],
    signing_secret=os.environ["SLACK_SIGNING_SECRET"],
)

# The default behavior won't be changed at all
assert response.status == 404
assert response.body == '{"error": "unhandled request"}'

app = App(
    token=os.environ["SLACK_BOT_TOKEN"],
    signing_secret=os.environ["SLACK_SIGNING_SECRET"],
    # enable @app.error handler to catch the patterns
    raise_error_for_unhandled_request=True,
)

@app.error
def handle_errors(error):
    if isinstance(error, BoltUnhandledRequestError):
        return BoltResponse(status=503, body="TODO")
    else:
        # other error patterns
        pass

assert response.status == 503
assert response.body == "TODO"
```

See the added unit tests for verifying the behavior.

* [ ] Others

### DIFF
--- a/slack_bolt/app/app.py
+++ b/slack_bolt/app/app.py
@@ -17,7 +17,7 @@ from slack_bolt.authorization.authorize import (
     InstallationStoreAuthorize,
     CallableAuthorize,
 )
-from slack_bolt.error import BoltError
+from slack_bolt.error import BoltError, BoltUnhandledRequestError
 from slack_bolt.lazy_listener.thread_runner import ThreadLazyListenerRunner
 from slack_bolt.listener.builtins import TokenRevocationListeners
 from slack_bolt.listener.custom_listener import CustomListener
@@ -50,6 +50,7 @@ from slack_bolt.logger.messages import (
     debug_return_listener_middleware_response,
     info_default_oauth_settings_loaded,
     error_installation_store_required_for_builtin_listeners,
+    warning_unhandled_by_global_middleware,
 )
 from slack_bolt.middleware import (
     Middleware,
@@ -85,6 +86,8 @@ class App:
         name: Optional[str] = None,
         # Set True when you run this app on a FaaS platform
         process_before_response: bool = False,
+        # Set True if you want to handle an unhandled request as an exception
+        raise_error_for_unhandled_request: bool = False,
         # Basic Information > Credentials > Signing Secret
         signing_secret: Optional[str] = None,
         # for single-workspace apps
@@ -94,7 +97,7 @@ class App:
         # for multi-workspace apps
         authorize: Optional[Callable[..., AuthorizeResult]] = None,
         installation_store: Optional[InstallationStore] = None,
-        # for v1.0.x compatibility
+        # for either only bot scope usage or v1.0.x compatibility
         installation_store_bot_only: Optional[bool] = None,
         # for the OAuth flow
         oauth_settings: Optional[OAuthSettings] = None,
@@ -132,6 +135,9 @@ class App:
             logger: The custom logger that can be used in this app.
             name: The application name that will be used in logging. If absent, the source file name will be used.
             process_before_response: True if this app runs on Function as a Service. (Default: False)
+            raise_error_for_unhandled_request: True if you want to raise exceptions for unhandled requests
+                and use @app.error listeners instead of
+                the built-in handler, which pints warning logs and returns 404 to Slack (Default: False)
             signing_secret: The Signing Secret value used for verifying requests from Slack.
             token: The bot/user access token required only for single-workspace app.
             token_verification_enabled: Verifies the validity of the given token if True.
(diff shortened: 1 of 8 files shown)


seratch:
#### Open questions to reviewers

* Can we agree to have the new option? Or may we want to introduce new listener name (say, `@app.unhandled` or `@app.unmatched`)?
* If we go with having a flag option, does the naming `raise_error_for_unhandled_request` look good to you?

slack_bolt/app/app.py
```diff
@@ -83,6 +84,8 @@ def __init__(
         name: Optional[str] = None,
         # Set True when you run this app on a FaaS platform
         process_before_response: bool = False,
+        # Set True if you want to handle an unhandled request as an exception
+        raise_error_for_unhandled_request: bool = False,
```
seratch, line 88: The term `raise` is a bit Python-specific. If we introduce the same option to JavaScript and Java, using `throw` for them is a reasonable option.

Another option may be `unhandled_request_error_enabled`, which can be programming language agnostic.

slack_bolt/listener/listener_error_handler.py
```diff
@@ -31,7 +31,7 @@ def handle(
 
 
 class CustomListenerErrorHandler(ListenerErrorHandler):
-    def __init__(self, logger: Logger, func: Callable[..., None]):
+    def __init__(self, logger: Logger, func: Callable[..., Optional[BoltResponse]]):
```
seratch, line 34: Customizing a response to Slack had not been supported. Actually, there is no obstacle or strong reason for that.

tests/scenario_tests/test_error_handler.py
```diff
@@ -153,3 +154,91 @@ def failing_listener():
...
+            raise_error_for_unhandled_request=True,
+        )
+        response = app.dispatch(self.build_valid_request())
+        assert response.status == 404
+        assert response.body == '{"error": "unhandled request"}'
+
+        @app.error
+        def handle_errors(error):
+            assert isinstance(error, BoltUnhandledRequestError)
+            return BoltResponse(status=404, body="TODO")
```
mwbrooks, line 171: Just curious, about how this test would fail. Let's pretend that there was a bug causing `raise_error_for_unhandled_request=True` to not raise the error. In that case, `handle_errors` would not be triggered and the assertion inside would not run.

In other test frameworks, we can "expect 4 asserts" to guarantee to catch a situation where callback pay not be triggered.

How does it work with Python in this case? I'm concerned that the assertion on L170 may not trigger but the test will pass.

mwbrooks (approved):
Nice work! 👍🏻  This would certainly be a nice addition to `bolt-js` and interesting to compare the two PRs for their differences.

tests/scenario_tests/test_error_handler.py
seratch: @mwbrooks 
>In that case, handle_errors would not be triggered and the assertion inside would not run.

L175 verifies if the response body is "TODO". As the default body is '{"error": "unhandled request"}', the assertion fails if this listener is not executed.

>In other test frameworks, we can "expect 4 asserts" to guarantee to catch a situation where callback pay not be triggered.

If the assertion in the `handle_errors` method fails, it raises an exception. As a result, the Bolt app responds with status 500. 

Regarding the other assertions in the test method, there is nothin special. If any of them fails, the test suite fails too.

PR #323, opened 2021-05-05
Add missing listener suggestion to the default unhandled error message

seratch:
This pull request improves the default warning message for unhandled requests by printing suggested code snippets as below:

```
$ python examples/socket_mode.py
⚡️ Bolt app is running!

Unhandled request ({'token': 'xxx', 'team_id': 'T111', 'api_app_id': 'A111', 'event': {'client_msg_id': 'xxx', 'type': 'message', 'text': 'Hi there!', 'user': 'U222', 'ts': '1620202537.001600', 'team': 'T111', 'channel': 'C111', 'event_ts': '1620202537.001600', 'channel_type': 'channel'}, 'type': 'event_callback', 'event_id': 'Ev111', 'event_time': 1620202537, 'authorizations': [{'enterprise_id': None, 'team_id': 'T111', 'user_id': 'U222', 'is_bot': True, 'is_enterprise_install': False}], 'is_ext_shared_channel': False, 'event_context': '1-message-T111-C111'})
---
[Suggestion] You can handle this type of event with the following listener function:

@app.event("message")
def handle_message_events(body, logger):
    logger.info(body)
```

With this suggestion, developers, particularly the ones new to bolt-python, can easily understand what to do for handling the requests in their app. This pull request covers all the possible patterns with corresponding code snippets. Refer to the added unit test code (`tests/slack_bolt/logger/test_unmatched_suggestions.py`) to learn the actual suggestion messages.

If a developer does not want to have the message, we would recommend having `raise_error_for_unhandled_request=True` in the `App`/`AsyncApp` constructor arguments and having their own global error handler (see "Customize Unhandled Error Handling" in the [v1.5.0 release notes](https://github.com/slackapi/bolt-python/releases/tag/v1.5.0) for more details).

* [ ] Others

### DIFF
--- a/scripts/run_tests.sh
+++ b/scripts/run_tests.sh
@@ -12,13 +12,13 @@ python_version=`python --version | awk '{print $2}'`
 if [[ $test_target != "" ]]
 then
   black slack_bolt/ tests/ && \
-    pytest $1
+    pytest -vv $1
 else
   if [ ${python_version:0:3} == "3.8" ]
   then
     # pytype's behavior can be different in older Python versions
     black slack_bolt/ tests/ \
-      && pytest \
+      && pytest -vv \
       && pip install -e ".[adapter]" \
       && pip install -U pip setuptools wheel \
       && pip install -U pytype \
--- a/slack_bolt/app/app.py
+++ b/slack_bolt/app/app.py
@@ -530,8 +530,6 @@ class App:
     def _handle_unmatched_requests(
         self, req: BoltRequest, resp: BoltResponse
     ) -> BoltResponse:
-        # TODO: provide more info like suggestion of listeners
-        # e.g., You can handle this type of message with @app.event("app_mention")
         self._framework_logger.warning(warning_unhandled_request(req))
         return resp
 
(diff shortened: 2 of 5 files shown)


stevengill (approved):
LGTM. I really like that you are showing these awesome code samples

Only question I have is that we are logging `req.body` for unhandled requests. In Bolt-js, usually we have stayed away from logging this in case folks are piping logs to some system and don't want sensitive information such as `token` in their logs. Are you concerned about that?

seratch:
@stevengill 
>Only question I have is that we are logging req.body for unhandled requests. In Bolt-js, usually we have stayed away from logging this in case folks are piping logs to some system and don't want sensitive information such as token in their logs. Are you concerned about that?

Thanks, this is a good point. Basically, I don't think logging request payloads is a problem as payloads do not have any access tokens and credentials. Payloads can include [the legacy verification token (we don't recommend using anymore and Bolt relies on only signature verification)](https://api.slack.com/authentication/verifying-requests-from-slack#verification_token_deprecation) and confidential data in text messages. I don't think we call these data as credentials. For this reason, I think logging the payload in debug-level logging does not have any issues as we already do in Bolt for JS's Developer Mode.

With that being said, this logging is warning-level and it's not easy to omit only the log if a developer has to eliminate the possibility of including confidential information in product app logs. I will adjust the log message for this warning-level log and will add another debug level log for knowing the whole payload (in the development phase). Thanks for the great input 👍

Issue #332, opened 2021-05-09
AWS Lambda adapter is not compatible with AWS event data format v1.0


matteobaldelli:
I am not an expert of AWS lambda but I'm trying to make my first lambda function with bolt but I think this line is wrong

https://github.com/slackapi/bolt-python/blob/31436f729871c38f596a1363aa1d61a5b4c15d0d/slack_bolt/adapter/aws_lambda/handler.py#L81

`cookies: Sequence[str] = event.get("cookies", [])` cookies will be always `[]` because with AWS cookies are not inside event but inside headers, so I think you should do something like that:
`cookies: Sequence[str] = event.get("headers", {}).get("Cookie", [])`

seratch:
Hi @matteobaldelli, thanks for writing in!

The AWS Lambda adapter supports only payload format v2.0. As you can see at [AWS's Developer Guide](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html), the `cookies` exists in the v2.0 format. Perhaps, your app should be using v1.0.

If we receive many requests for supporting both 1.0 and 2.0, we may update the handler code to be compatible with both.

matteobaldelli:
Hi @seratch, I did not know about different version of payload, thanks for the information!

I checked the documentation and now I understand what I was doing wrong with my lambda, maybe you are already aware of it but I still want describe the problem that could be useful to others.
I am using API Gateway to launch my lambda, when you create a new endpoint with API Gateway, you can choose between REST API and HTTP API, I had chosen REST API.
The problem is that REST API don't support payload format v2.0, so since the adapter not support payload format v1.0 I can't use REST API for hit my lambda.
I migrated my endpoint from REST API to HTTP API and now everything works, if you use payload format v1.0 the oauth_flow pages in GET (`/slack/install` and `/slack/oauth_redirect`) will not work.
I opened this issues for the same problem https://github.com/slackapi/bolt-python/issues/330

Maybe I miss that but will be useful to write in the documentation to use payload format v2.0 or support both.

Thanks for the support!

seratch:
@matteobaldelli Thanks a lot for sharing your knowledge with others! 

[...]

Issue #322, opened 2021-05-02
best way of URL handling


tomerbsentinel:
I wonder if there is a "best practice" or everyone is building their own URL parsing mechanism.
Slack are formatting the text into hyperlinks, and I need to overcome this.

I'm getting URLs like: (depends who entered it to the text input)
only text -  google.com
short URL - <google.com|google.com>
Long URL - \<https://google.com|google.com>
 
1. I didn't see it, but, do we have the exact input of the end-user
2. Do anyone have a good practice of how to get only the text out of the pre-formatted command?

seratch:
Hi @tomerbsentinel , thanks for asking the question!

>I didn't see it, but, do we have the exact input of the end-user

No, we don't. Only the top-level `text` property in a payload is available for apps.

>Do anyone have a good practice of how to get only the text out of the pre-formatted command?

We're still consulting other experts but, as far as I know, there is no standard way to do so. For your use case, the current recommendation would be writing your own parser. 🙇

seratch:
👋 I will close this issue soon. Let us know if you have something further to ask/discuss here!

PR #379, opened 2021-06-15
Make cookies extraction on AWS Lambda compatible with its format v1.0

tattee:
I tried to deploy a slack bolt application on AWS Lambda with SQLAlchemy.
In the process of the verification of oauth redirect, the app could not be installed properly.
I found that the slack request handler could not extract cookie from the lambda event.
So, I modified the to_bolt_request to meet the format of the lambda event.

### DIFF
--- a/slack_bolt/adapter/aws_lambda/handler.py
+++ b/slack_bolt/adapter/aws_lambda/handler.py
@@ -84,8 +84,10 @@ def to_bolt_request(event) -> BoltRequest:
     body = event.get("body", "")
     if event["isBase64Encoded"]:
         body = base64.b64decode(body).decode("utf-8")
-    multiValueHeaders = event.get("multiValueHeaders", {})
-    cookies: Sequence[str] = multiValueHeaders.get("Cookie", [])
+    cookies: Sequence[str] = event.get("cookies", [])
+    if event.get("requestContext", {}).get("http", {}).get("method") is None: # judging whether payload format version is v1.0 or not
+        multiValueHeaders = event.get("multiValueHeaders", {})
+        cookies = multiValueHeaders.get("Cookie", [])
     headers = event.get("headers", {})
     headers["cookie"] = cookies
     return BoltRequest(
--- a/slack_bolt/adapter/aws_lambda/handler.py
+++ b/slack_bolt/adapter/aws_lambda/handler.py
@@ -84,7 +84,8 @@ def to_bolt_request(event) -> BoltRequest:
     body = event.get("body", "")
     if event["isBase64Encoded"]:
         body = base64.b64decode(body).decode("utf-8")
-    cookies: Sequence[str] = event.get("cookies", [])
+    multiValueHeaders = event.get("multiValueHeaders", {})
+    cookies: Sequence[str] = multiValueHeaders.get("Cookie", [])
     headers = event.get("headers", {})
     headers["cookie"] = cookies
     return BoltRequest(
--- a/slack_bolt/adapter/aws_lambda/handler.py
+++ b/slack_bolt/adapter/aws_lambda/handler.py
@@ -85,7 +85,8 @@ def to_bolt_request(event) -> BoltRequest:
     if event["isBase64Encoded"]:
         body = base64.b64decode(body).decode("utf-8")
     cookies: Sequence[str] = event.get("cookies", [])
-    if event.get("requestContext", {}).get("http", {}).get("method") is None: # judging whether payload format version is v1.0 or not
+    if cookies is None or len(cookies) == 0:
+        # In the case of format v1
         multiValueHeaders = event.get("multiValueHeaders", {})
         cookies = multiValueHeaders.get("Cookie", [])
     headers = event.get("headers", {})
--- a/tests/adapter_tests/aws/test_aws_lambda.py
+++ b/tests/adapter_tests/aws/test_aws_lambda.py
@@ -318,3 +318,41 @@ class TestAWSLambda:
         assert response["statusCode"] == 200
         assert response["headers"]["content-type"] == "text/html; charset=utf-8"
         assert "https://slack.com/oauth/v2/authorize?state=" in response.get("body")
+
+    @mock_lambda
+    def test_oauth_redirect(self):
+        app = App(
+            client=self.web_client,
+            signing_secret=self.signing_secret,
+            oauth_settings=OAuthSettings(
+                client_id="111.111",
+                client_secret="xxx",
+                scopes=["chat:write", "commands"],
+            ),
+        )
+
+        event = {
+            "body": "",
+            "queryStringParameters": {"code": "1234567890", "state": "uuid4-value"},
+            "headers": {},
+            "cookies": ["slack-app-oauth-state=uuid4-value"],
+            "requestContext": {"http": {"method": "GET"}},
+            "isBase64Encoded": False,
+        }
+        response = SlackRequestHandler(app).handle(event, self.context)
+        assert response["statusCode"] == 401
+        assert response["headers"]["content-type"] == "text/html; charset=utf-8"
+        assert response.get("body") is not None
+
+        event = {
+            "body": "",
+            "queryStringParameters": {"code": "1234567890", "state": "uuid4-value"},
+            "headers": {},
+            "multiValueHeaders": {"Cookie": ["slack-app-oauth-state=uuid4-value"]},
+            "requestContext": {"httpMethod": "GET"},
+            "isBase64Encoded": False,
+        }
+        response = SlackRequestHandler(app).handle(event, self.context)
+        assert response["statusCode"] == 401
+        assert response["headers"]["content-type"] == "text/html; charset=utf-8"
+        assert response.get("body") is not None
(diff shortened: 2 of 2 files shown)


seratch:
Hi @tattee, thank you very much for takin the time to make this pull request. Before my code review, I have two things to ask:

* Can you associate the email address that you use for git commits with your GitHub account?
* And then, would you mind signing our CLA (see the above message by the CLA bot)?

Even if your changes are great, we are unable to merge them without signed CLA. I would appreciate it if you could understand this.

seratch:
HI @tattee, thanks for taking the time to make this pull request. Can you check this issue https://github.com/slackapi/bolt-python/issues/332#issuecomment-839587227 ? This project supports only AWS lambda event payload format v2.0 as the HTTP API is the currently recommended way plus it is the greatest fit for running Bolt apps.

With that being said, we are open to support format v1.0 as well. If you have more time, could you help us add the support by making the following additional work?

* Add unit tests verifying if it works with v1.0 format
  * We already have tests with v2.0 format [here](https://github.com/slackapi/bolt-python/blob/main/tests/adapter_tests/aws/test_aws_lambda.py), you can add more tests in the same file
  * The tests should have a realistic payload
* Update the document after merging the change (I can do this later)

tattee:
@seratch thank you for telling me the issue #322.
I looked at the [unit test code for aws lambda handler](https://github.com/slackapi/bolt-python/blob/main/tests/adapter_tests/aws/test_aws_lambda.py) and found that the test function named "[test_oauth](https://github.com/slackapi/bolt-python/blob/0927de45b265bc4d16570ceb6a23cb5dd0a21317/tests/adapter_tests/aws/test_aws_lambda.py#L287)" only tests the process of outh_flow.handle_installation().
So, I added new test method named "[test_oauth_redirect](https://github.com/slackapi/bolt-python/pull/379/commits/06711f9704b3c027aa09e1a144d77ffed5375d4b)".
The response from SlackRequestHandler only returns success or failure of authentication, and it does not help understand whether urlStringParameters are extracted properly.
Fortunately, in the process of outh_flow.handle_callback(), the verification of the value of "state" is taken place(in [oauth_flow.py](https://github.com/slackapi/bolt-python/blob/0927de45b265bc4d16570ceb6a23cb5dd0a21317/slack_bolt/oauth/oauth_flow.py#L218)).
If the value of state is correct, the statusCode of reponse changes from 400 to 401.
Although the authentication does not succeed, you can verify that the parameters in the event were read correctly by checking that the status code value changes to 401.

If my test code is OK, I will sent a pull request again.

seratch:
@tattee Thanks for taking the time to work on it. Ideally, we should have some tests for `/slack/oauth_redirect` handling but the test does not exist yet. Thus, we don't need to add a new test pattern in this pull request. We can improve it next time.

My previous comment might not be clear enough but the tests we need this time is the same patterns with the ones for the already supported AWS payload format. Considering the number of existing test methods, adding the same to the same file may not be great. How about having `test_aws_lambda_format_v1.py` and `test_aws_lambda_format_v2.py`? Having two files should be far easier to understand the intention and maintain.

tattee:
@seratch In my understanding, the format of the events in [test_oauth](https://github.com/slackapi/bolt-python/blob/0927de45b265bc4d16570ceb6a23cb5dd0a21317/tests/adapter_tests/aws/test_aws_lambda.py#L287) are v2.0 (first one) and v1.0 (second one) respectively, and the processes in the tests are almost same.
I'm afraid dividing the test file would increase mental burden. 

If my concern does not make sense, please ignore it.

seratch:
Ah thanks. I totally forgot these tests already support both formats. So, this means that only cookies extraction is not yet fully supported for both formats, right? In the case, trying to check `/slack/oauth_redirect` pattern is reasonable. I'm sorry about my confusion. I will add a few comments to the changes.

tests/adapter_tests/aws/test_aws_lambda.py
```diff
@@ -318,3 +318,41 @@ def test_oauth(self):
         assert response["statusCode"] == 200
         assert response["headers"]["content-type"] == "text/html; charset=utf-8"
         assert "https://slack.com/oauth/v2/authorize?state=" in response.get("body")
+
+    @mock_lambda
+    def test_oauth_redirect(self):
+        app = App(
```
seratch, line 324: If we have a mock state_store, which returns a fixed value "uuid4-value", the test can be better. Can you check this?

slack_bolt/adapter/aws_lambda/handler.py
```diff
@@ -85,6 +85,9 @@ def to_bolt_request(event) -> BoltRequest:
     if event["isBase64Encoded"]:
         body = base64.b64decode(body).decode("utf-8")
     cookies: Sequence[str] = event.get("cookies", [])
+    if event.get("requestContext", {}).get("http", {}).get("method") is None: # judging whether payload format version is v1.0 or not
```
seratch, line 88:
```suggestion
    if cookies is None or len(cookies) == 0:
        # In the case of format v1
```

tattee (commented):
I checked your comments, and left replies.
Best regards,

tests/adapter_tests/aws/test_aws_lambda.py
seratch: @tattee Thanks for asking this! I thought that we need to implement state_store with `issue`/`consume` methods (returning a fixed value "uuid4-value" means having the issue method). But, for this test, having only `consume` method is required. Thus, the following diff should work for you!

```diff
diff --git a/tests/adapter_tests/aws/test_aws_lambda.py b/tests/adapter_tests/aws/test_aws_lambda.py
index b92300e..1b0166c 100644
--- a/tests/adapter_tests/aws/test_aws_lambda.py
+++ b/tests/adapter_tests/aws/test_aws_lambda.py
@@ -3,6 +3,7 @@ from time import time
 from urllib.parse import quote
 
 from moto import mock_lambda
+from slack_sdk.oauth import OAuthStateStore
 from slack_sdk.signature import SignatureVerifier
 from slack_sdk.web import WebClient
 
@@ -321,6 +322,10 @@ class TestAWSLambda:
 
     @mock_lambda
     def test_oauth_redirect(self):
+        class TestStateStore(OAuthStateStore):
+            def consume(self, state: str) -> bool:
+                return state == "uuid4-value"
+
         app = App(
             client=self.web_client,
             signing_secret=self.signing_secret,
@@ -328,6 +333,7 @@ class TestAWSLambda:
                 client_id="111.111",
                 client_secret="xxx",
                 scopes=["chat:write", "commands"],
+                state_store=TestStateStore(),
             ),
         )
 
@@ -340,7 +346,7 @@ class TestAWSLambda:
             "isBase64Encoded": False,
         }
         response = SlackRequestHandler(app).handle(event, self.context)
-        assert response["statusCode"] == 401
+        assert response["statusCode"] == 200
         assert response["headers"]["content-type"] == "text/html; charset=utf-8"
         assert response.get("body") is not None
 
@@ -353,6 +359,6 @@ class TestAWSLambda:
             "isBase64Encoded": False,
         }
         response = SlackRequestHandler(app).handle(event, self.context)
-        assert response["statusCode"] == 401
+        assert response["statusCode"] == 200
         assert response["headers"]["content-type"] == "text/html; charset=utf-8"
         assert response.get("body") is not None
```

seratch (commented):
Thanks for updating this PR. Once the test is improved, we can merge your PR 👍

tattee:
It's a nice idea to add TestStateStore for consume method.
I reflected your diff on test file.
(Sorry, I don't know how to execute pytest, and I did not confirm the test succeeds.)
If your diff is enough, I will submit new commit. If not, could you tell me how to use pytest?
Best regards,

seratch:
@tattee Thanks for the comment! You can run pytest for this project [this way](https://github.com/slackapi/bolt-python/blob/main/.github/workflows/ci-build.yml). Also, there is a [tiny script](https://github.com/slackapi/bolt-python/blob/main/scripts/run_tests.sh) too.

tattee:
Thank you for telling me how to execute test scripts.
I confirmed the test succeeded.
I pushed commit reflecting your test_oauth_redirect method.
Sorry for the late submission.

seratch (approved):
LGTM - thanks!

weallwegot:
this is great- thank you! i realize now that this closes the loop on this [exchange we had a while ago](https://github.com/slackapi/bolt-python/commit/7da0f7c0644836fe3485f7d8d6cb8a78af7d0993#r48492950) around v1 vs v2 support, glad to see this PR & release

naveensan1:
@seratch I too deployed a Bolt-python app to AWS Lambda, integrated with REST (format v1.0) based API GW. In the process of the verification of oauth redirect, the app could not be installed properly. Taking a look at the CloudWatch logs I could see the multivalueheader in the request had the key "cookie" instead of "Cookie" as is expected by "to_bolt_request".

```python
   if cookies is None or len(cookies) == 0:
        # In the case of format v1
        multiValueHeaders = event.get("multiValueHeaders", {})
        cookies = multiValueHeaders.get("Cookie", [])
```

https://github.com/slackapi/bolt-python/blob/main/slack_bolt/adapter/aws_lambda/handler.py#L91

After modifying the following line to account for lower case "cookie", everything worked fine:

```python
     cookies = multiValueHeaders.get("cookie", [])
```
Recommend this "get" operation be modified to be case-insensitive as it is blocking oauth install at the moment.

seratch:
@naveensan1 Thanks for sharing this! This makes sense 👍 If you have a chance, would you mind sending a pull request to fix this?

PR #394, opened 2021-07-07
Fix #370 by adding an alias of `next` arg (`next_`) in middleware arguments

### LINKED ISSUE #370: The `next` in middleware args overwrites Python built-in one
### `slack_bolt` version

Current [`HEAD`](https://github.com/slackapi/bolt-python/blob/66aad210faeed75c04941dfb635a4556952a8868/slack_bolt/middleware/middleware.py).

### Description
The keyword `next` used in `slack_bolt.middleware.middleware.process()` overwrites the [Python builtin `next()`](https://docs.python.org/3/library/functions.html#next), used to produce the next item in iterables and generators. We are currently trying to pass a generator to a custom middleware, having us resorting in using a hack to use the generator as middleware.

Please consider using another variable name, or `next_`.

Cheers, and thank you for your great work! ❤️

C.C. @nikodraca

seratch:
This pull request fixes #370. The essential change in this pull request is to add `next_` to kwargs_injection names. With this fix, developers can use `next_` in addition to `next` in the middleware method arguments.

```python
@app.middleware
def just_call_next(next_):
    next_()
```

* [ ] Others

### DIFF
--- a/slack_bolt/kwargs_injection/utils.py
+++ b/slack_bolt/kwargs_injection/utils.py
@@ -52,6 +52,7 @@ def build_required_kwargs(
         "respond": request.context.respond,
         # middleware
         "next": next_func,
+        "next_": next_func,  # for the middleware using Python's built-in `next()` function
     }
     all_available_args["payload"] = (
         all_available_args["options"]
--- a/slack_bolt/listener/listener.py
+++ b/slack_bolt/listener/listener.py
@@ -45,10 +45,10 @@ class Listener(metaclass=ABCMeta):
         for m in self.middleware:
             middleware_state = {"next_called": False}
 
-            def next():
+            def next_():
                 middleware_state["next_called"] = True
 
-            resp = m.process(req=req, resp=resp, next=next)
+            resp = m.process(req=req, resp=resp, next=next_)
             if not middleware_state["next_called"]:
                 # next() was not called in this middleware
                 return (resp, True)
--- a/tests/scenario_tests/test_middleware.py
+++ b/tests/scenario_tests/test_middleware.py
@@ -87,6 +87,53 @@ class TestMiddleware:
         assert response.body == "acknowledged!"
         assert_auth_test_count(self, 1)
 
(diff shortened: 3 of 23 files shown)


tests/scenario_tests/test_middleware.py
```diff
@@ -87,6 +87,53 @@ def test_next_call(self):
         assert response.body == "acknowledged!"
         assert_auth_test_count(self, 1)
 
+    def test_decorator_next_call(self):
```
seratch, line 90: Checking these test patterns should be helpful for knowing the purpose of the changes.

slack_bolt/listener/listener.py
```diff
@@ -45,10 +45,10 @@ def run_middleware(
         for m in self.middleware:
             middleware_state = {"next_called": False}
 
-            def next():
+            def next_():
```
seratch, line 48: This is not an essential part of this PR. It's just a minor refactoring

slack_bolt/kwargs_injection/utils.py
```diff
@@ -52,6 +52,7 @@ def build_required_kwargs(
         "respond": request.context.respond,
         # middleware
         "next": next_func,
+        "next_": next_func,  # for the middleware using Python's built-in `next()` function
```
seratch, line 55: Adding this enables bolt-python to resolve the alias name as well.

seratch (commented):
comments for reviewers

mwbrooks (approved):
Looks good to me! ✅ 

Just out of curiosity, is `next_` a convention in Python? Rather than say `nextMiddleware`.

seratch:
>Just out of curiosity, is next_ a convention in Python? Rather than say nextMiddleware.

I think `next_middleware` is also fine but it may be a bit long (in comparison with other middleware args). As we use `for` syntax (`for element in seq:`) in most cases, it's rare to directly call `next` with an iterator. Having a kind of workaround name like this should make sense.

PR #453, opened 2021-08-28
Allow a custom Executor in App

### LINKED ISSUE #452: Allow App's listener_executor to be supplied to __init__
It would be helpful to be able to supply my own executor to my Bolt app. For my particular use-case, I would like to use an executor from [Flask-Executor](https://pypi.org/project/Flask-Executor/) for an app running in Flask, so that my listener functions automatically run with the current request context.

Additionally, it would allow users to configure the number of threads and the threading strategy (min threads, max threads, etc.).

I would be willing to create a pull request for this feature if it's acceptable. I suspect all that would be required is a new keyword-only parameter to `App.__init__`. We could fall back to the existing behavior if it's `None`.

My current work-around is to monkey-patch the executors in `ThreadListenerRunner` and `ThreadLazyListenerRunner`. Not difficult, but not future-proof.

* [ ] **slack_bolt.async_app.AsyncApp** and/or its core components
* [ ] Others

## Requirements

chrisbouchard:
This pull request adds a `listener_executor` keyword parameter to `App.__init__`, so that the executor for running background listener functions can be customized by the user. It also weakens the type requirement for executors to just `concurrent.futures.Executor`, the abstract base class of executors.

Fixes #452.

* [ ] Others

### DIFF
--- a/slack_bolt/app/app.py
+++ b/slack_bolt/app/app.py
@@ -175,7 +175,8 @@ class App:
             oauth_settings: The settings related to Slack app installation flow (OAuth flow)
             oauth_flow: Instantiated `slack_bolt.oauth.OAuthFlow`. This is always prioritized over oauth_settings.
             verification_token: Deprecated verification mechanism. This can used only for ssl_check requests.
-            listener_executor: Optional executor to run background tasks. If absent, a ThreadPoolExecutor will be used.
+            listener_executor: Optional executor to run background tasks. If absent, a `ThreadPoolExecutor` will
+                be used.
         """
         signing_secret = signing_secret or os.environ.get("SLACK_SIGNING_SECRET")
         token = token or os.environ.get("SLACK_BOT_TOKEN")
--- a/slack_bolt/app/app.py
+++ b/slack_bolt/app/app.py
@@ -114,6 +114,7 @@ class App:
         oauth_flow: Optional[OAuthFlow] = None,
         # No need to set (the value is used only in response to ssl_check requests)
         verification_token: Optional[str] = None,
+        # Set this one only when you want to customize the executor
         listener_executor: Optional[Executor] = None,
     ):
         """Bolt App that provides functionalities to register middleware/listeners.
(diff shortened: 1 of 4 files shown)


chrisbouchard:
I think this is ready for review. One thing I'm unsure of is whether I should update the API documentation. I did update the docstring on `App.__init__`. When I ran `./scripts/generate_api_docs.sh` it updated way more files than I touched.

mwbrooks (commented):
Hey @chrisbouchard, thanks for following up with a pull request! You're awesome 🎉 

The CLA looks good and I've enabled the test workflow to run. ✅ 

I'll let @seratch review your pull request and comment on why generating the documentation changes more files than expected. Regardless, thanks a lot for taking the extra steps to add test and update the docs 🤘🏻

chrisbouchard:
Ok, I was able to figure out the issue with docs. It looks like a new version of pydoc has been published (0.10.0) that causes content and formatting changes in around 171 doc files. I hard-coded my pydoc version to 0.9.3 to match what's currently checked into main and I got a much more reasonable diff.

seratch:
@chrisbouchard Thanks for updating the docs but can you revert the change? We usually update the API docs after releasing a version.

slack_bolt/app/app.py
```diff
@@ -173,6 +175,8 @@ def message_hello(message, say):
             oauth_settings: The settings related to Slack app installation flow (OAuth flow)
             oauth_flow: Instantiated `slack_bolt.oauth.OAuthFlow`. This is always prioritized over oauth_settings.
             verification_token: Deprecated verification mechanism. This can used only for ssl_check requests.
+            listener_executor: Optional executor to run background tasks. If absent, a `ThreadPoolExecutor` will
```
seratch, line 178: As the executor is not optional (some executor is required), I prefer "custom" for describing the given one 
```suggestion
            listener_executor: Custom executor to run background tasks. If absent, the default `ThreadPoolExecutor` will
```

slack_bolt/app/app.py
```diff
@@ -113,6 +114,7 @@ def __init__(
         oauth_flow: Optional[OAuthFlow] = None,
         # No need to set (the value is used only in response to ssl_check requests)
         verification_token: Optional[str] = None,
+        listener_executor: Optional[Executor] = None,
```
seratch, line 117:
```suggestion
        # Set this one only when you want to customize the executor
        listener_executor: Optional[Executor] = None,
```

seratch (changes requested):

chrisbouchard:
Doc changes reverted, and wording suggestions applied. :+1:

tests/scenario_tests/test_app.py
```diff
@@ -56,6 +58,22 @@ def test_listener_registration_error(self):
         with pytest.raises(BoltError):
             app.action({"type": "invalid_type", "action_id": "a"})(self.simple_listener)
 
+    def test_listener_executor(self):
+        class TestExecutor(Executor):
+            """A stupid executor that does nothing."""
```
seratch, line 63:
```suggestion
            """A executor that does nothing for testing"""
```

seratch (approved):

seratch:
Thanks for working on this! LGTM

PR #281, opened 2021-04-02
Fix #280 Django thread-local connection cleanup in multi threads

seratch:
This pull request resolves #280 . For the context and details of the issue we're going to address, refer to the issue #280. I will leave a few comments for reviewers.

### Category (place an `x` in each of the `[ ]`)

* [x] `slack_bolt.App` and/or its core components
* [x] `slack_bolt.async_app.AsyncApp` and/or its core components
* [x] Adapters in `slack_bolt.adapter`
* [ ] Document pages under `/docs`
* [ ] Others

## Requirements (place an `x` in each `[ ]`)

Please read the [Contributing guidelines](https://github.com/slackapi/bolt-python/blob/main/.github/contributing.md) and [Code of Conduct](https://slackhq.github.io/code-of-conduct) before creating this issue or pull request. By submitting, you are agreeing to those rules.

* [x] I've read and understood the [Contributing Guidelines](https://github.com/slackapi/bolt-python/blob/main/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).
* [x] I've run `./scripts/install_all_and_run_tests.sh` after making the changes.

Issue #280, opened 2021-04-02
Django thread-local connection cleanup in multi threads 


seratch:
This issue illustrates a potential issues that may affect Django adapter users.

Django's ORM establishes a database connection per thread. For the threads managed by Django (which are supposed to be used for directly handing web requests), Django ORM cleans the established connections up under the hood once the processing on the thread completes (= request_finished signal).

On the other hand, in the case where Django ORM models are used in an unmanaged thread, Django does not handle the cleanup of the connections bound to the thread although the Django framework automatically associates a new connection to the thread. This half-heartedly maintained resource possibly may cause stale DB connection related issues.

To learn more details of this Django behavior, checking the following ticket is helpful for understanding this:
https://code.djangoproject.com/ticket/9878

As Bolt for Python utilizes threads for enabling developers to run `ack()` method asynchronously (as long as `process_before_response=False`, which is the default), this framework should have a special treatment for this possible issue with Django.

As the solution for this issue (and possible similar needs), we'll introduce a new completion callback to the `ListenerRunner` mechanism. The handler will be quite similar to the `ListenerErrorHandler` callback but the given callback function is executed in the `finally` clause of listener runner invocation: 
https://github.com/slackapi/bolt-python/blob/v1.4.4/slack_bolt/listener/thread_runner.py#L99-L126
Also, for this Django specific issue, we'll add a custom lazy listener runner to the Django adapter module.

 I've already implemented the initial version of this. We can discuss its details in my upcoming pull request.

PR #509, opened 2021-11-01
Use django.db close_old_connections in place of explicit connections.close

ross:
I was looking to enable the django db `CONN_MAX_AGE` to avoid the overhead of reconnecting to the database (TLS encrypted) for every event being handled similar to how I have in the past with normal django web requests. 

I quickly realized this wasn't working as I continued to see the mysql connection counters on the server rising for each request. I dug into the log message about closing connections and ran across `release_thread_local_connections`  https://github.com/slackapi/bolt-python/blob/99ed6a6c1bb09c8f186b916eaf54e963073a112c/slack_bolt/adapter/django/handler.py#L60

That method appears to explicitly close all connections after each event is handled which makes `CONN_MAX_AGE` ineffective. 

This PR makes a relatively minor change to `release_thread_local_connections` to instead call the same `request_finished` handler the signal mentioned in https://github.com/slackapi/bolt-python/issues/280#issue-849076180, `django.db.close_old_connections` which appears to be easy/safe since it ignores all args with `**kwargs`. 

This opens things up to connection reuse via `CONN_MAX_AGE` and otherwise gets the full django connection cleanup behavior on errors instead of explicitly closing the connection for each event handled.

/cc https://github.com/slackapi/bolt-python/issues/280 https://github.com/slackapi/bolt-python/pull/281 which introduced the explicit close
/cc @seratch who authored ^ and would likely know if this makes sense

* [ ] Others

### DIFF
--- a/slack_bolt/adapter/django/handler.py
+++ b/slack_bolt/adapter/django/handler.py
@@ -54,15 +54,15 @@ def to_django_response(bolt_resp: BoltResponse) -> HttpResponse:
     return resp
 
 
-from django.db import connections
+from django.db import close_old_connections
 
 
 def release_thread_local_connections(logger: Logger, execution_type: str):
-    connections.close_all()
+    close_old_connections()
     if logger.level <= logging.DEBUG:
         current: Thread = current_thread()
         logger.debug(
-            f"Released thread-bound DB connections (thread name: {current.name}, execution type: {execution_type})"
+            f"close_old_connections called to manage thread-bound DB connections (thread name: {current.name}, execution type: {execution_type})"
         )
 
 
@@ -124,8 +124,8 @@ class SlackRequestHandler:
             Bolt skipped to set it to slack_sdk.adapter.django.DjangoListenerCompletionHandler.
             We strongly recommend having the following lines of code in your listener_completion_handler:
 
-            from django.db import connections
-            connections.close_all()
+            from django.db import close_old_connections
+            close_old_connections()
             """
             self.app.logger.warning(message)
             return


ross:
> Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/slackapi/bolt-python?pullRequest=509) before we can accept your contribution.

I don't sign legal documents as part of OSS contributions. [CLA aren't necessary](https://about.gitlab.com/blog/2017/11/01/gitlab-switches-to-dco-license/) and are bad for contributors. Very few true OSS projects require them, I've only really seen them on company sponsored projects like this. [They're bad practice](https://drewdevault.com/2018/10/05/Dont-sign-a-CLA.html).

Feel free to port/redo the trivial changes in this PR yourself as you see fit.

seratch:
Hi @ross, thanks a lot for taking the time to share this feedback and a suggestion to resolve the issue!

I've checked `close_old_connections` and I totally agree that using the method is far better than the current approach. One thing I noticed is that, if we switch to the method, perhaps we may want to call the method before middleware/listener execution as well (or only before the execution). Refer to https://github.com/django/channels/issues/1227 for details. Some Slack apps may not receive incoming requests for a few hours / days. It's not rare. For this reason, doing the same would be necessary for safety.

>Feel free to port/redo the trivial changes in this PR yourself as you see fit.

Regarding the CLA, we respect your opinion about it and thank you very much for saying this. We'll work on a new pull request based on your suggestion. If you are happy to use your time a little bit more, your reviews on our forthcoming pull request would be appreciated.

ross:
> I've checked `close_old_connections` and I totally agree that using the method is far better than the current approach. One thing I noticed is that, if we switch to the method, perhaps we may want to call the method before middleware/listener execution as well (or only before the execution). Refer to [django/channels#1227](https://github.com/django/channels/issues/1227) for details. Some Slack apps may not receive incoming requests for a few hours / days. It's not rare. For this reason, doing the same would be necessary for safety.

Thanks. I was totally focused on the closing aspect and didn't even look at the start signal enough to see that it was calling the method as well. Definitely makes sense to do it similarly for slack_bolt even handlers.

> Regarding the CLA, we respect your opinion about it and thank you very much for saying this. We'll work on a new pull request based on your suggestion. If you are happy to use your time a little bit more, your reviews on our forthcoming pull request would be appreciated.

Wfm. Happy to take a look at and test PRs. Thanks.

ross:
Not directly related to this PR/issue, but after failing to find any information about using Django from `Thread`s or `ThreadPoolExecutor`s in searches I decided to create a python module `django_thread` with simple derivatives of those classes that take care of the connection management. If you all wanted you could switch to using `django_thread.ThreadPoolExecutor` to get the correct pre and post `close_old_connections` calls wrapped around job execution transparently. I know the places you've hooked in here are part of the wider framework so that may make sense to just continue doing that.

seratch:
@ross Thanks for sharing this! Your solution for this Django issue is very clean and looks great to me 👍  If your idea will be included out-of-the-box in the Django framework in the future (=encouraging a Django specific thread executor for threading), it should be so helpful not only for this Slack app use case but also for the entire community.

ross:
Closed in favor of https://github.com/slackapi/bolt-python/pull/512

PR #512, opened 2021-11-02
Improve the Django DB connection management #509

seratch:
This pull request improves the internals of Django adapter. To better maintain the database connections bound to the threads created by bolt, not Django. Refer to the following issues and the comments that I've added in this PR for more details:

* https://github.com/slackapi/bolt-python/pull/509
* https://github.com/slackapi/bolt-python/issues/280
* https://github.com/slackapi/bolt-python/pull/281

I'm so grateful for @ross's help at #509. Thank you very much for helping us improve the adapter.

* [ ] Others

### DIFF
--- a/slack_bolt/adapter/django/handler.py
+++ b/slack_bolt/adapter/django/handler.py
@@ -54,15 +54,16 @@ def to_django_response(bolt_resp: BoltResponse) -> HttpResponse:
     return resp
 
 
-from django.db import connections
+from django.db import close_old_connections
 
 
-def release_thread_local_connections(logger: Logger, execution_type: str):
-    connections.close_all()
+def release_thread_local_connections(logger: Logger, execution_timing: str):
+    close_old_connections()
     if logger.level <= logging.DEBUG:
         current: Thread = current_thread()
         logger.debug(
-            f"Released thread-bound DB connections (thread name: {current.name}, execution type: {execution_type})"
+            "Released thread-bound old DB connections "
+            f"(thread name: {current.name}, execution timing: {execution_timing})"
         )
 
 
@@ -73,7 +74,7 @@ class DjangoListenerCompletionHandler(ListenerCompletionHandler):
     """
 
     def handle(self, request: BoltRequest, response: Optional[BoltResponse]) -> None:
-        release_thread_local_connections(request.context.logger, "listener")
+        release_thread_local_connections(request.context.logger, "listener-completion")
 
 
 class DjangoThreadLazyListenerRunner(ThreadLazyListenerRunner):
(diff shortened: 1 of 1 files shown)


ross:
I'm seeing 3 calls to `close_old_connections()` from 3 different threads. Is that expected? 

I tried to trace through the code around this change, but I'm not familiar enough with it to easily suss out what all the Runner and Handler stuff is doing.

My assumption here is that `close_old_connections()` needs to be called on the `ThreadPoolExecutor` thread that's handling the request. Once just before firing off the event and then again after the event has been handled. 

### First

This appears first thing as soon as i've triggered the event in slack.

```
Thread-1
...
  File ".../slacker/listeners.py", line 125, in slack_events_handler
    return handler.handle(request)
  File ".../env/src/slack-bolt/slack_bolt/adapter/django/handler.py", line 153, in handle
    release_thread_local_connections(
```

### Second

Around the time the app.event fires. This is the one that surprises me.

```
ThreadPoolExecutor-2_0
...
  File ".../env/src/slack-bolt/slack_bolt/listener/thread_runner.py", line 135, in run_ack_function_asynchronously
    self.listener_completion_handler.handle(
  File ".../env/src/slack-bolt/slack_bolt/adapter/django/handler.py", line 79, in handle
    release_thread_local_connections(request.context.logger, "listener-completion")
```

### Third

Final one is after my event handling has completed. 

```
Thread-2
...
  File ".../slacker/listeners.py", line 125, in slack_events_handler
    return handler.handle(request)
  File ".../env/src/slack-bolt/slack_bolt/adapter/django/handler.py", line 153, in handle
    release_thread_local_connections(
```

I've dumped the thread names and tracebacks by adding the following to the top of `release_thread_local_connections`

```python
    from traceback import print_stack
    current: Thread = current_thread()
    print(f'\n\n{current.name}')
    print_stack()
```

ross:
I assume the new calls are coming from https://github.com/slackapi/bolt-python/pull/512/files#diff-910415999171fcdec47153808936ea3a8c30d2a1b52f71e0cea137a3fe5de7d4R151, but still not sure why there's 2 of them.  

Was expecting one before the app.event is fired to do the lifetime checks as you mentioned over in https://github.com/slackapi/bolt-python/pull/509#issuecomment-955879763. End up deciding against that?

seratch:
@ross Thanks for your input here. 

>Was expecting one before the app.event is fired to do the lifetime checks as you mentioned over in #509 (comment). End up deciding against that?

`app.event` listeners can be invoked inside the `app.dispatch` method. Thus, this pull request should added a beforehand `close_old_connections` call in some way. 

But the point that I might overlook is that it's done in the Django's request handing thread. Perhaps, the timing and the thread to run the operation may need to be changed. As you pointed out, I'm thinking that the beforehand call of `close_old_connections` should be in the threads managed by Bolt.

PR #514, opened 2021-11-03
Introduce ListenerStartHandler in the listener runner for better managing Django DB connections

seratch:
Long story short, this pull request is a revised solution of #512 

This patch can resolve the potential resource issue with Django ORM. The ORM automatically acquires a database connection when performing database queries in a thread but if the thread is not managed by Django Web framework, such connections won't be properly released.

In the pull request https://github.com/slackapi/bolt-python/pull/512, I was wanting to resolve this issue by changing only the internals of the Django adapter. However, thanks to @ross's great feedback https://github.com/slackapi/bolt-python/pull/512#issuecomment-958344238, we found that the approach does not work as expected. The clean up method calls must be done in the thread that is used for asynchronous listener execution.

To resolve this issue, we have to add a new hook point called `listener_start_handler` in the `ThreadListenerRunner`. We already have `listener_error_handler` and `listener_completion_handler`. In a similar way, we'll add a new handler that is supposed to be called at the beginning of `ThreadListenerRunner`. The handler can perform `django.db. close_old_connections()` method in the expected thread, plus before the listener execution.

In the sake of maintaining the feature parity in the asyncio-based `AsyncApp`, this pull request adds `AsyncListenerStartHandler` as well. But the asyncio-based ones are unrelated to the Django issue.

@ross 
Thank you very much for your help (again). If you have any comments on this pull request, your inputs would be greatly appreciated.

* [ ] Others

### DIFF
--- a/slack_bolt/version.py
+++ b/slack_bolt/version.py
@@ -1,2 +1,2 @@
 """Check the latest version at https://pypi.org/project/slack-bolt/"""
-__version__ = "1.9.4"
+__version__ = "1.10.0a"
--- a/slack_bolt/adapter/django/handler.py
+++ b/slack_bolt/adapter/django/handler.py
@@ -9,6 +9,10 @@ from slack_bolt.app import App
 from slack_bolt.error import BoltError
 from slack_bolt.lazy_listener import ThreadLazyListenerRunner
 from slack_bolt.lazy_listener.internals import build_runnable_function
+from slack_bolt.listener.listener_start_handler import (
+    ListenerStartHandler,
+    DefaultListenerStartHandler,
+)
 from slack_bolt.listener.listener_completion_handler import (
     ListenerCompletionHandler,
     DefaultListenerCompletionHandler,
@@ -67,6 +71,16 @@ def release_thread_local_connections(logger: Logger, execution_timing: str):
         )
 
 
+class DjangoListenerStartHandler(ListenerStartHandler):
+    """Django sets DB connections as a thread-local variable per thread.
+    If the thread is not managed on the Django app side, the connections won't be released by Django.
+    This handler releases the connections every time a ThreadListenerRunner execution completes.
+    """
+
+    def handle(self, request: BoltRequest, response: Optional[BoltResponse]) -> None:
+        release_thread_local_connections(request.context.logger, "listener-start")
+
+
 class DjangoListenerCompletionHandler(ListenerCompletionHandler):
     """Django sets DB connections as a thread-local variable per thread.
     If the thread is not managed on the Django app side, the connections won't be released by Django.
@@ -86,6 +100,9 @@ class DjangoThreadLazyListenerRunner(ThreadLazyListenerRunner):
         )
 
         def wrapped_func():
+            release_thread_local_connections(
+                request.context.logger, "before-lazy-listener"
+            )
             try:
                 func()
             finally:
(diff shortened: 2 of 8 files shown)


slack_bolt/listener/async_listener_completion_handler.py
```diff
@@ -15,10 +15,9 @@ async def handle(
         request: AsyncBoltRequest,
         response: Optional[BoltResponse],
     ) -> None:
-        """Handles an unhandled exception.
```
seratch, line 18: This had been wrong 🤦

slack_bolt/listener/listener_completion_handler.py
```diff
@@ -15,7 +15,7 @@ def handle(
         request: BoltRequest,
         response: Optional[BoltResponse],
     ) -> None:
-        """Handles an unhandled exception.
+        """Do something extra after the listener execution
```
seratch, line 18: This had been wrong 🤦

slack_bolt/version.py
```diff
@@ -1,2 +1,2 @@
 """Check the latest version at https://pypi.org/project/slack-bolt/"""
-__version__ = "1.9.4"
+__version__ = "1.10.0a"
```
seratch, line 2: We'll bump the minor version as this pull request introduces new functionalities

filmaj (commented):
... given I approved the last PR that led to the thrice-execution issue described by @ross, not sure my review should count for much 🙈 😓 

I need to first familiarize myself with Django + Bolt 🤐

ross:
I'm not seeing two calls fro the right thread with the expected `timing`s. Still doing a bit of testing to make sure they're at the expected points.

```
2021-11-03T15:59:50Z DEBUG Released thread-bound old DB connections (thread name: bot-worker_0, execution timing: listener-start)
...
2021-11-03T15:59:51Z DEBUG Released thread-bound old DB connections (thread name: bot-worker_0, execution timing: listener-completion)
```

The `ThreadPoolExecutor` being created when one is not provided is using the default names on the worker threads. https://github.com/slackapi/bolt-python/blob/cf3951e4eebe7f72daf235d285023fbda090d24a/slack_bolt/app/app.py#L311

Adding `thread_name_prefix=slack_bolt-workers` or similar would give those threads better names to more easily identify them when digging into things like this. I'm passing my own `ThreadPoolExecutor` in above which is why mine have names.

ross:
Verified that the calls are wrapping the `app.event` call. So I think we're good there. I also went ahead and verified the calls happening to `close_old_connections` including threads and I think they're as expected:

```
2021-11-03T16:13:45Z INFO  db close_old_connections: thread=Thread-1
...
2021-11-03T16:13:46Z INFO  db close_old_connections: thread=simone-worker_0
...
2021-11-03T16:13:46Z INFO  db close_old_connections: thread=Thread-1
...
2021-11-03T16:13:48Z INFO  db close_old_connections: thread=simone-worker_0
```

Based on stack traces the `Thread-1` connections are coming from django's `runserver` and are the http threads not owned by `slack_bolt` so noise w/respect to this PR.

There's a 3rd pair that happens just after that. It appears to be coming from the self message triggered by my bot's reply so that might have been the source of the `Thread-2` second one I couldn't account for in https://github.com/slackapi/bolt-python/pull/512#issuecomment-958610514.

I can't do a formal approval or anything so 👍 will have to count :grin:

seratch:
@ross @filmaj Thanks a lot for quickly checking this PR!

seratch:
I will make a new minor version release later today

Issue #606, opened 2022-02-28
Options loaded from a url are not filtering on characters typed


r4ibOm:
Hello,

I created a modal and an item in the modal is a selection input, of the type "external_select" that looks for a list of options.

I did it as described in the documentation and the list with all the items appears when I click on the listbox. The problem is that when the user starts typing, instead of filtering by the characters typed, it is always showing all the options.

``` python
from slack_utils.main import app


@app.shortcut("requisicao_id")
def handle_shortcuts(ack, body, client):
    # Acknowledge the command request
    ack()
    # Call views_open with the built-in client
    client.views_open(
        # Pass a valid trigger_id within 3 seconds of receiving it
        trigger_id=body["trigger_id"],
        # View payload
        view={
            "type": "modal",
            # View identifier
            "callback_id": "view_1",
            "title": {"type": "plain_text", "text": "My App"},
            "submit": {"type": "plain_text", "text": "Submit"},
            "blocks": [
                {
                    "type": "section",
                    "text": {"type": "mrkdwn", "text": "Welcome to a modal with _blocks_"},
                    "accessory": {
                        "type": "button",
                        "text": {"type": "plain_text", "text": "Click me!"},
                        "action_id": "button_abc"
                    }
                },
                {
                    "type": "input",
                    "block_id": "input_c",
                    "label": {"type": "plain_text", "text": "What are your hopes and dreams?"},
                    "element": {
                        "type": "plain_text_input",
                        "action_id": "dreamy_input",
                        "multiline": True
                    }
                },
                {
                    "type": "input",
                    "element": {
                        "action_id": "aws_access_select",
                        "type": "external_select",
                        "placeholder": {
                            "type": "plain_text",
                            "text": "Select or enter the expected value..."
                        },
                        "min_query_length": 0
                    },
                    "label": {
                        "type": "plain_text",
                        "text": "Label",
                        "emoji": True
                    }
                }
            ]
        }
    )


@app.view("view_1")
def handle_view_events(ack, body, client, view, logger):
    # Assume there's an input block with `input_c` as the block_id and `dreamy_input`
    hopes_and_dreams = view["state"]["values"]["input_c"]["dreamy_input"]["value"]
    user = body["user"]["id"]
    # Validate the inputs
    errors = {}
    if hopes_and_dreams is not None and len(hopes_and_dreams) <= 5:
        errors["input_c"] = "The value must be longer than 5 characters"
    if len(errors) > 0:
[...]

PR #609, opened 2022-03-01
Docs on handling options listeners with a filtering example

filmaj:
Relates to #606 

Once again did my best with Japanese translations but Google Translate takes credit for that - could use help with that side of things @seratch @wongjas !

### DIFF
--- a/docs/_basic/ja_listening_responding_options.md
+++ b/docs/_basic/ja_listening_responding_options.md
@@ -14,7 +14,7 @@ order: 14
 
 オプションのリクエストに応答するときは、有効なオプションを含む `options` または `option_groups` のリストとともに `ack()` を呼び出す必要があります。API サイトにある[外部データを使用する選択メニューに応答するサンプル例](https://api.slack.com/reference/messaging/block-elements#external-select)と、[ダイアログでの応答例](https://api.slack.com/dialogs#dynamic_select_elements_external)を参考にしてください。
 
-さらに、ユーザー入力に基づいて、返されたオプションにフィルタリングロジックを適用することもできます。 これは、オプションリスナーへの `payload`引数を使用し、その中の` value`プロパティの内容をチェックすることで実現できます。 `value`に基づいて、さまざまなオプションを返すことができます。 すべてのBoltPythonリスナーとミドルウェアハンドラーは、多くの有用な引数にアクセスできます。必ずチェックしてください。
+さらに、ユーザーが入力したキーワードに基づいたオプションを返すようフィルタリングロジックを適用することもできます。 これは `payload` という引数の ` value` の値に基づいて、それぞれのパターンで異なるオプションの一覧を返すように実装することができます。 Bolt for Python のすべてのリスナーやミドルウェアでは、[多くの有用な引数](https://slack.dev/bolt-python/api-docs/slack_bolt/kwargs_injection/args.html)にアクセスすることができますので、チェックしてみてください。
 
 </div>
 
--- a/docs/_basic/ja_listening_responding_options.md
+++ b/docs/_basic/ja_listening_responding_options.md
@@ -14,6 +14,8 @@ order: 14
 
 オプションのリクエストに応答するときは、有効なオプションを含む `options` または `option_groups` のリストとともに `ack()` を呼び出す必要があります。API サイトにある[外部データを使用する選択メニューに応答するサンプル例](https://api.slack.com/reference/messaging/block-elements#external-select)と、[ダイアログでの応答例](https://api.slack.com/dialogs#dynamic_select_elements_external)を参考にしてください。
 
+さらに、ユーザー入力に基づいて、返されたオプションにフィルタリングロジックを適用することもできます。 これは、オプションリスナーへの `payload`引数を使用し、その中の` value`プロパティの内容をチェックすることで実現できます。 `value`に基づいて、さまざまなオプションを返すことができます。 すべてのBoltPythonリスナーとミドルウェアハンドラーは、多くの有用な引数にアクセスできます。必ずチェックしてください。
+
 </div>
 
 <div>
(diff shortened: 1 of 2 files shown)


srajiang (commented):
Since this is the section about handling options, would it be more straightforward in the code sample just to use the `options` kwarg directly instead of the body and refer to the `payload` and `body` as alternatives which provide access to other data from the Slack incoming request? What `payload` and `body` contain (and some of the unstated redundancy) can confuse people.

filmaj:
Hmm, I'm not sure updating the same to use the `options` argument would actually help. Since the listener method is supposed to acknowledge with an array of options, I think one could get easily confused between the `options` argument provided to the listener (which contain things like the specific input value typed in by the end-user) to the `options` needing to acknowledge the listener invocation with (which contains a list of options to feed to the dropdown). IMO it is clearer to use `body` or `payload` and keep the variable name inside the handler as `options`.

srajiang:
>  Since the listener method is supposed to acknowledge with an array of options, I think one could get easily confused between the options argument provided to the listener (which contain things like the specific input value typed in by the end-user) to the options needing to acknowledge the listener invocation with (which contains a list of options to feed to the dropdown).
That's valid, though that lack of clarity you point out is why I think it could be documented better. Perhaps `payload` would be better to use, since `options` is an alias of `body` and body contains other request body data that isn't needed for this here.

filmaj:
That's fair, let me take another stab at writing that up in a clearer way.

seratch:
@srajiang @filmaj I agree that we can use `payload` rather than `body`. I feel `options` can be confusing in the code as we pass `options` to the `ack` method.

In the case of block_suggestion handling, actually, `body` and `payload` are exactly the same. If we suggest using `body`, the use case should be the situation where a developer's code needs access to top-level properties such as `body.trigger_id` in view_submission requests.

filmaj:
OK I've tried to rewrite it based on your comments. Let me know what you think.

srajiang:
Ah, so `body` `payload` and `options` are all the same in this case? TIL From the kwargs docs it seemed as though `options` and `payload` were aliases and `body` was somehow different from the others. Perhaps that is true in other cases? How confusing.

srajiang (approved):
English looks good to me!

docs/_basic/ja_listening_responding_options.md
```diff
@@ -14,24 +14,29 @@ order: 14
 
 オプションのリクエストに応答するときは、有効なオプションを含む `options` または `option_groups` のリストとともに `ack()` を呼び出す必要があります。API サイトにある[外部データを使用する選択メニューに応答するサンプル例](https://api.slack.com/reference/messaging/block-elements#external-select)と、[ダイアログでの応答例](https://api.slack.com/dialogs#dynamic_select_elements_external)を参考にしてください。
 
+さらに、ユーザー入力に基づいて、返されたオプションにフィルタリングロジックを適用することもできます。 これは、オプションリスナーへの `payload`引数を使用し、その中の` value`プロパティの内容をチェックすることで実現できます。 `value`に基づいて、さまざまなオプションを返すことができます。 すべてのBoltPythonリスナーとミドルウェアハンドラーは、多くの有用な引数にアクセスできます。必ずチェックしてください。
```
seratch, line 17:
```suggestion
さらに、ユーザーが入力したキーワードに基づいたオプションを返すようフィルタリングロジックを適用することもできます。 これは `payload` という引数の ` value` の値に基づいて、それぞれのパターンで異なるオプションの一覧を返すように実装することができます。 Bolt for Python のすべてのリスナーやミドルウェアでは、[多くの有用な引数](https://slack.dev/bolt-python/api-docs/slack_bolt/kwargs_injection/args.html)にアクセスすることができますので、チェックしてみてください。
```

docs/_basic/listening_responding_options.md
```diff
@@ -13,14 +13,15 @@ While it's recommended to use `action_id` for `external_select` menus, dialogs d
 
 To respond to options requests, you'll need to call `ack()` with a valid `options` or `option_groups` list. Both [external select response examples](https://api.slack.com/reference/messaging/block-elements#external-select) and [dialog response examples](https://api.slack.com/dialogs#dynamic_select_elements_external) can be found on our API site.
 
+Additionally, you may want to apply filtering logic to the returned options based on user input. This can be accomplished by using the `payload` argument to your options listener and checking for the contents of the `value` property within it. Based on the `value` you can return different options. All Bolt Python listeners and middleware handlers have access to [many useful arguments](https://slack.dev/bolt-python/api-docs/slack_bolt/kwargs_injection/args.html) - be sure to check them out!
```
seratch, line 16: "Bolt Python" is not a term that we publicly use. How about saying as below instead?
```suggestion
Additionally, you may want to apply filtering logic to the returned options based on user input. This can be accomplished by using the `payload` argument to your options listener and checking for the contents of the `value` property within it. Based on the `value` you can return different options. All listeners and middleware handlers in Bolt for Python have access to [many useful arguments](https://slack.dev/bolt-python/api-docs/slack_bolt/kwargs_injection/args.html) - be sure to check them out!
```

seratch:
@srajiang 
>From the kwargs docs it seemed as though options and payload were aliases and body was somehow different from the others. Is that true in all cases? How confusing.

This a bit unusual situation can arise for block_suggestion.

The intention why Bolt has `payload` in addition to `body` is to enable developers to easily extract the essential part of the whole payload (say, `body.actions[0]` in block_actions, `body.view` in view_submission). For block_suggestion, the top-level ones are already the essential part of the payload. This behavior is indeed confusing but that depends on how the request data looks like. Bolt tries to simplify users' code when it's possible.

filmaj:
Thanks for the reviews @srajiang and @seratch ! I've committed your suggestions. If you think it is good, please merge!

seratch (approved):

PR #614, opened 2022-03-15
Add ASGI Falcon App support

sarayourfriend:
This PR adds an ASGI Falcon App adapter to complement the existing Falcon adapter for WSGI apps.

I wrote this code originally based on the existing adapter for https://git.sr.ht/~sara/openverse-slack-reaction.

The code I wrote there is copied almost directly here with some modifications to raise errors when an incompatible Falcon version is being used. You can see the original code here: https://git.sr.ht/~sara/openverse-slack-reaction/tree/main/item/falcon_adapter.py

It's deployed to production with https://openverse-slack.sarayourfriend.pictures.

Hope it's okay I didn't open an issue for this first, but I needed it for my own app one way or another and copying the file  and adding tests was trivial so I didn't spend too much extra time on this compared to just writing a helpful issue describing the feature :slightly_smiling_face: 

* [ ] Others

### DIFF
--- a/examples/falcon/async_app.py
+++ b/examples/falcon/async_app.py
@@ -1,0 +1,100 @@
+import falcon
+import logging
+import re
+from slack_bolt.async_app import AsyncApp, AsyncRespond, AsyncAck
+from slack_bolt.adapter.falcon import AsyncSlackAppResource
+
+logging.basicConfig(level=logging.DEBUG)
+app = AsyncApp()
+
+
+# @app.command("/bolt-py-proto", [lambda body: body["team_id"] == "T03E94MJU"])
+async def test_command(logger: logging.Logger, body: dict, ack: AsyncAck, respond: AsyncRespond):
+    logger.info(body)
+    await ack("thanks!")
+    await respond(
+        blocks=[
+            {
+                "type": "section",
+                "block_id": "b",
+                "text": {
+                    "type": "mrkdwn",
+                    "text": "You can add a button alongside text in your message. ",
+                },
+                "accessory": {
+                    "type": "button",
+                    "action_id": "a",
+                    "text": {"type": "plain_text", "text": "Button"},
+                    "value": "click_me_123",
+                },
+            }
+        ]
+    )
+
+
+app.command(re.compile(r"/hello-bolt-.+"))(test_command)
+
+
+@app.shortcut("test-shortcut")
+async def test_shortcut(ack, client, logger, body):
+    logger.info(body)
+    await ack()
+    res = await client.views_open(
+        trigger_id=body["trigger_id"],
+        view={
+            "type": "modal",
+            "callback_id": "view-id",
+            "title": {
+                "type": "plain_text",
+                "text": "My App",
+            },
+            "submit": {
+                "type": "plain_text",
+                "text": "Submit",
+            },
+            "close": {
+                "type": "plain_text",
+                "text": "Cancel",
+            },
+            "blocks": [
+                {
+                    "type": "input",
+                    "element": {"type": "plain_text_input"},
+                    "label": {
+                        "type": "plain_text",
+                        "text": "Label",
+                    },
+                }
+            ],
+        },
+    )
+    logger.info(res)
+
+
+@app.view("view-id")
+async def view_submission(ack, body, logger):
+    logger.info(body)
+    await ack()
+
+
+@app.action("a")
+async def button_click(logger, action, ack, respond):
+    logger.info(action)
+    await ack()
+    await respond("Here is my response")
+
+
+@app.event("app_mention")
+async def handle_app_mentions(body, say, logger):
+    logger.info(body)
+    await say("What's up?")
+
+
+api = falcon.asgi.App()
+resource = AsyncSlackAppResource(app)
+api.add_route("/slack/events", resource)
+
+# pip install -r requirements.txt
+# export SLACK_SIGNING_SECRET=***
+# export SLACK_BOT_TOKEN=xoxb-***
+# uvicorn --reload -h 0.0.0.0 -p 3000 app:api
--- a/examples/falcon/async_app.py
+++ b/examples/falcon/async_app.py
@@ -97,4 +97,4 @@ api.add_route("/slack/events", resource)
 # pip install -r requirements.txt
 # export SLACK_SIGNING_SECRET=***
 # export SLACK_BOT_TOKEN=xoxb-***
-# uvicorn --reload -h 0.0.0.0 -p 3000 app:api
+# uvicorn --reload -h 0.0.0.0 -p 3000 async_app:api
--- a/examples/falcon/async_oauth_app.py
+++ b/examples/falcon/async_oauth_app.py
@@ -97,6 +97,6 @@ api.add_route("/slack/events", resource)
 # pip install -r requirements.txt
 # export SLACK_SIGNING_SECRET=***
 # export SLACK_BOT_TOKEN=xoxb-***
-# uvicorn --reload -h 0.0.0.0 -p 3000 app:api
+# uvicorn --reload -h 0.0.0.0 -p 3000 async_oauth_app:api
 api.add_route("/slack/install", resource)
 api.add_route("/slack/oauth_redirect", resource)
--- a/examples/falcon/async_oauth_app.py
+++ b/examples/falcon/async_oauth_app.py
@@ -1,0 +1,102 @@
+import falcon
+import logging
+import re
+from slack_bolt.async_app import AsyncApp, AsyncRespond, AsyncAck
+from slack_bolt.adapter.falcon import AsyncSlackAppResource
+
+logging.basicConfig(level=logging.DEBUG)
+app = AsyncApp()
+
+
+# @app.command("/bolt-py-proto", [lambda body: body["team_id"] == "T03E94MJU"])
+async def test_command(logger: logging.Logger, body: dict, ack: AsyncAck, respond: AsyncRespond):
+    logger.info(body)
+    await ack("thanks!")
+    await respond(
+        blocks=[
+            {
+                "type": "section",
+                "block_id": "b",
+                "text": {
+                    "type": "mrkdwn",
+                    "text": "You can add a button alongside text in your message. ",
+                },
+                "accessory": {
+                    "type": "button",
+                    "action_id": "a",
+                    "text": {"type": "plain_text", "text": "Button"},
+                    "value": "click_me_123",
+                },
+            }
+        ]
+    )
+
+
+app.command(re.compile(r"/hello-bolt-.+"))(test_command)
+
+
+@app.shortcut("test-shortcut")
+async def test_shortcut(ack, client, logger, body):
+    logger.info(body)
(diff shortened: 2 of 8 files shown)


slack_bolt/adapter/falcon/__init__.py
```diff
@@ -1 +1,2 @@
 from .resource import SlackAppResource
+from .async_resource import AsyncSlackAppResource
```
seratch, line 2:
```suggestion
# Don't add async module imports here
from .resource import SlackAppResource
```

slack_bolt/adapter/falcon/async_resource.py
```diff
@@ -0,0 +1,84 @@
...
+
+    async def _to_bolt_request(self, req: Request) -> AsyncBoltRequest:
+        return AsyncBoltRequest(
+            body=(await req.stream.read(req.content_length or 0)).decode("utf-8"),
+            query=req.query_string,
+            headers={k.lower(): v for k, v in req.headers.items()},
+        )
+
+    async def _write_response(self, bolt_resp: BoltResponse, resp: Response):
+        if falcon_version.__version__.startswith("2."):
```
seratch, line 59: Can we move this validation in the constructor instead?

The if/else in the sync adapter is added to deal with the Falcon's API changes in major versions (`resp.body` in v2 vs `resp.text` in v3). As this ASGI version supports only v3, we can simply have `resp.text = bolt_resp.body` only in this adapter.

That being said, this version check is still good for better developer experience! I think that it can be done earlier than the timing that your app receive requests from Slack.

slack_bolt/adapter/falcon/async_resource.py
```diff
@@ -0,0 +1,84 @@
...
+            raise NotImplementedError("ASGI Falcon requires version >= 3.0")
```
seratch, line 60:
```suggestion
            raise BoltError("This ASGI compatible adapter requires Falcon version >= 3.0")
```

tests/adapter_tests_async/test_async_falcon.py
```diff
@@ -0,0 +1,212 @@
...
+from tests.mock_web_api_server import (
+    setup_mock_web_api_server,
+    cleanup_mock_web_api_server,
+    assert_auth_test_count,
+)
+from tests.utils import remove_os_env_temporarily, restore_os_env
+
+
+def new_falcon_app():
+    if falcon.version.__version__.startswith("2."):
```
seratch, line 23: You can remove this if condition

seratch (changes requested):
Thanks for the great addition! We are happy to have this enhancement but could you check my comments and update your branch?

slack_bolt/adapter/falcon/__init__.py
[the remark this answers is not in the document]
sarayourfriend: Interesting, that hadn't crossed my mind, thanks for catching it!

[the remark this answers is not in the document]
sarayourfriend: Done in https://github.com/slackapi/bolt-python/pull/614/commits/937fddc20c28c2f467a71f1a849b4287b0ab362d

slack_bolt/adapter/falcon/async_resource.py
sarayourfriend: Done in https://github.com/slackapi/bolt-python/pull/614/commits/515345c42aa1ca12034650f5d231829d4d27adf5

tests/adapter_tests_async/test_async_falcon.py
sarayourfriend: Done in https://github.com/slackapi/bolt-python/pull/614/commits/515345c42aa1ca12034650f5d231829d4d27adf5

sarayourfriend:
Thanks for the quick review @seratch! I've updated the PR with your requested changes.

examples/falcon/async_app.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+api = falcon.asgi.App()
+resource = AsyncSlackAppResource(app)
+api.add_route("/slack/events", resource)
+
+# pip install -r requirements.txt
+# export SLACK_SIGNING_SECRET=***
+# export SLACK_BOT_TOKEN=xoxb-***
+# uvicorn --reload -h 0.0.0.0 -p 3000 app:api
```
seratch, line 100:
```suggestion
# uvicorn --reload -h 0.0.0.0 -p 3000 async_app:api
```

examples/falcon/async_oauth_app.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+api = falcon.asgi.App()
+resource = AsyncSlackAppResource(app)
+api.add_route("/slack/events", resource)
+
+# pip install -r requirements.txt
+# export SLACK_SIGNING_SECRET=***
+# export SLACK_BOT_TOKEN=xoxb-***
+# uvicorn --reload -h 0.0.0.0 -p 3000 app:api
```
seratch, line 100:
```suggestion
# uvicorn --reload -h 0.0.0.0 -p 3000 async_oauth_app:api
```

seratch (changes requested):
Can you update a few comments in the example apps?

examples/falcon/async_app.py
seratch: @sarayourfriend Also, I would suggest gunicorn here because the requirements.txt has only gunicorn

seratch:
@sarayourfriend Thanks for quick updates on the feedback! My last request is to update the examples a bit. Once the builds pass, we can merge this PR and release a new version by the end of this week.

sarayourfriend:
Hmm, it seems like there's an issue with the fact that we have tests that expect Falcon 2 instead of 3... is there some way to exclude the async Falcon tests when the falcon version is `2.x`?

examples/falcon/async_app.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+api = falcon.asgi.App()
+resource = AsyncSlackAppResource(app)
+api.add_route("/slack/events", resource)
+
+# pip install -r requirements.txt
+# export SLACK_SIGNING_SECRET=***
+# export SLACK_BOT_TOKEN=xoxb-***
+# uvicorn --reload -h 0.0.0.0 -p 3000 async_app:api
```
seratch, line 100: @sarayourfriend Can you add unvicorn to requirements.txt in the same directory?

seratch:
@sarayourfriend You can change this line to use `>=3,<4` instead: https://github.com/slackapi/bolt-python/blob/v1.11.6/setup.py#L82
Changing the part does not affect users.

sarayourfriend:
@seratch should I also remove the conditional in the existing Falcon adapter tests? I've never worked with `setup.py` before so just want to make sure I'm understanding correctly that it will run the tests only with Falcon 3.x?

It could make this function unnecessary: https://github.com/sarayourfriend/bolt-python/blob/78b64878b692039e39530888a012bf738ca74058/tests/adapter_tests/falcon/test_falcon.py#L23

examples/falcon/async_app.py
sarayourfriend: Sure thing. I'm not sure what version range to pin it to. The project is still <1. Should I just use `uvicorn>=0.17.6` (the currently released version)?

seratch:
@sarayourfriend Ah, sorry - never mind my above comment. I totally forget what I did in the past 🤦 No need to modify `setup.py`. 

Can you modify https://github.com/slackapi/bolt-python/blob/v1.11.6/.github/workflows/tests.yml#L73 to have `pip install "falcon>=3,<4"` before the pytest execution? It should resolve the issue.

sarayourfriend:
@seratch just wanted to make sure, because I don't see any other explicit dependency installations in that workflow... would it make sense to add the `falcon>=3` range in the `async` dependency list as that's already being installed in the `tests.yml`

This is the patch I would apply if that makes sense:

```diff
diff --git a/setup.py b/setup.py
index 6543529..9c92d94 100755
--- a/setup.py
+++ b/setup.py
@@ -68,6 +68,8 @@ setuptools.setup(
             "aiohttp>=3,<4",
             # Socket Mode 3rd party implementation
             "websockets>=8,<10",
+            # Falcon introduced ASGI support in v3
+            "falcon>=3,<4",
         ],
         # pip install -e ".[adapter]"
         # NOTE: any of async ones requires pip install -e ".[async]" too
@@ -79,7 +81,7 @@ setuptools.setup(
             "click>=7,<8",  # for chalice
             "CherryPy>=18,<19",
             "Django>=3,<5",
-            "falcon>=3,<4",
+            "falcon>=2,<4",
             "fastapi>=0.70.0,<1",
             "Flask>=1,<3",
             "Werkzeug>=2,<3",
```

If this doesn't work for some reason or you still prefer the explicit installation in the `tests.yml` script then let me know and I'll make the appropriate change.

seratch:
@sarayourfriend this line manually installs v2: https://github.com/slackapi/bolt-python/blob/v1.11.6/.github/workflows/tests.yml#L56 this is necessary for v2 tests but it causes errors in the async tests you added.

examples/falcon/async_app.py
seratch: You can do the same with fastapi examples

sarayourfriend:
Ah, I see. Makes sense, thanks for explaining. Pushed the update!

setup.py
```diff
@@ -79,7 +79,7 @@
             "click>=7,<8",  # for chalice
             "CherryPy>=18,<19",
             "Django>=3,<5",
-            "falcon>=2,<4",
+            "falcon>=3,<4",
```
seratch, line 82: Can you revert this change? This is not necessary
```suggestion
            "falcon>=2,<4",
```

slack_bolt/adapter/falcon/async_resource.py
```diff
@@ -0,0 +1,83 @@
...
+
+    async def _to_bolt_request(self, req: Request) -> AsyncBoltRequest:
+        return AsyncBoltRequest(
+            body=(await req.stream.read(req.content_length or 0)).decode("utf-8"),
+            query=req.query_string,
+            headers={k.lower(): v for k, v in req.headers.items()},
+        )
+
+    async def _write_response(self, bolt_resp: BoltResponse, resp: Response):
+        status = HTTPStatus(bolt_resp.status)
```
seratch, line 63: As you completely removed the part setting response body here, the unit test is correctly failing.
```suggestion
        resp.text = bolt_resp.body
        status = HTTPStatus(bolt_resp.status)
```

seratch (changes requested):
@sarayourfriend I found that your unit test properly detects a bug in your code. Can you check my comments?

slack_bolt/adapter/falcon/async_resource.py
sarayourfriend: Oh dear :cold_sweat: I was not careful in that change. Thanks for finding the fix.

setup.py
sarayourfriend: I wondered about this! :+1:

seratch (approved):
LGTM

PR #645, opened 2022-05-11
Fix #644 app.message listener does not handle events when a file is attached

### LINKED ISSUE #644: app.message listener does not handle events when a file is attached
When a message has attached files, the corresponding message event has file_share subtype. In this scenario, bolt-python's app.message listener does not handle the event even when the text matches. 

### Reproducible in:

any versions of bolt-python, any runtime

#### Steps to reproduce:

(Share the commands to run, source code, and project settings (e.g., setup.py))

1. Have app.message listener with keyword in your bolt-python app
2. Post a message that matches the keyword with any files

### Expected result:

The event should be handled by the app.message listener

### Actual result:

The event listener does not handle the event while it does if a message does not have files

seratch:
This pull request resolves #644 

When it comes to the feature parity with bolt-js, we may want to remove the subtype constraints in bolt-python. bolt-js does not check any subtypes when matching message text in `app.message` listeners: https://github.com/slackapi/bolt-js/blob/%40slack/bolt%403.11.0/src/App.ts#L602-L615

That being said, changing bolt-python app.message listeners to handle other message patterns by Slack product (e.g., [channel_join](https://api.slack.com/events/message/channel_join), [channel_posting_permissions](https://api.slack.com/events/message/channel_posting_permissions)) could be a breaking change to existing apps. Also, more importantly, passing [message_changed](https://api.slack.com/events/message/message_changed) and so on would require code changes on the bolt-python user side for sure plus the behavior should not be desired at all. 

For this reason, I didn't change the part drastically and went with just adding file_share subtype to the allowed list.

* [ ] Others

### DIFF
--- a/slack_bolt/app/app.py
+++ b/slack_bolt/app/app.py
@@ -837,6 +837,9 @@ class App:
                     # If an end-user posts a message with "Also send to #channel" checked,
                     # the message event comes with this subtype.
                     "thread_broadcast",
+                    # If an end-user posts a message with attached files,
+                    # the message event comes with this subtype.
+                    "file_share",
                 ),
             }
             primary_matcher = builtin_matchers.message_event(
--- a/slack_bolt/app/async_app.py
+++ b/slack_bolt/app/async_app.py
@@ -894,6 +894,9 @@ class AsyncApp:
                     # If an end-user posts a message with "Also send to #channel" checked,
                     # the message event comes with this subtype.
                     "thread_broadcast",
+                    # If an end-user posts a message with attached files,
+                    # the message event comes with this subtype.
+                    "file_share",
                 ),
             }
             primary_matcher = builtin_matchers.message_event(
--- a/tests/scenario_tests/test_message_file_share.py
+++ b/tests/scenario_tests/test_message_file_share.py
@@ -1,0 +1,173 @@
+import json
+import time
+
+from slack_sdk.signature import SignatureVerifier
+from slack_sdk.web import WebClient
+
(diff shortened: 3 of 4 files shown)


filmaj (approved):
Thanks for this change!

Regarding the potential breaking change to line up behaviour with bolt-js, my view on this is that consistency between bolt-js and bolt-python is more a nicety for us as maintainers rather than for bolt users. I think bolt users would probably stay with one language, generally, and therefore not expect identical behaviour if they move to a different language for bolt? What does everyone else think?

If the above assumption is accurate or agreed on by the rest of the team, then releasing a breaking change just to improve maintainers lives seems heavy-handed and should not be done just for that one change. However, if we have other, user-facing breaking changes we would like to land, then adding the subtype matching change in together with other breaking changes in the next major release may make sense.

filmaj:
Thank you for that additional context @celestinojones, your input helps make sure we are in line with customer expectations 🙏

seratch:
@filmaj @celestinojones Thanks for the comments! 

As for other subtypes apart from file_share, we may want to revisit in the future if it's worth. However, I think that we should not decide to simply change the app.message listener behavior without any options to opt-in (for existing bolt-python apps). I have to say that the subtypes in message events are tricky in the nature. We can be more careful than usual for the behavior change.

Either way, this file_share subtype pattern is an obvious bug, so that we should fix this as early as possible. Thus, we can merge this PR for the upcoming patch version release now. Thanks for the review and discussion!

Issue #646, opened 2022-05-12
403 Error when executing 'slack/install' in GCP cloud functions


ggm1207:
First of all, I apologize that I may not be able to write smoothly because my English skill is not good. I will write in as much detail as possible, so please ask again if there are any missing parts.

I'm implementing and debugging a slackbot in 2 environments. One is my Macbook and the other is GCP Cloud Functions.

As mentioned in the title, if I try to download(in workspace) the Slack bot using the url below, I will get a 403 Error (see the picture below). However, the problem is fine when I run Server using ngrok + flask on Macbook, but it only occurs when I run it in Cloud Functions.

- url(example): `{SERVER_ENDPOINT}/slack/install`

![image](https://user-images.githubusercontent.com/40788624/168141644-4007bedb-1242-4fed-bc98-5721d0a21242.png)

I'm not sure if this is an Oauth issue or a GCP permission issue or something.

#### The `slack_bolt` version

- slack-bolt==1.11.4
- slack-sdk==3.13.0

#### Python runtime version

- Python 3.9.1

#### OS info

- Cloud Functions
    - Can't provide
- Macbook
    - ProductName:	macOS
    - ProductVersion:	11.3.1
    - BuildVersion:	20E241
    - Darwin Kernel Version 20.4.0: Thu Apr 22 21:46:47 PDT 2021; root:xnu-7195.101.2~1/RELEASE_X86_64

#### Steps to reproduce:

1. Place the [sample file](https://gist.github.com/ggm1207/6389641e8e7403d49babbfdd9b032e9d) somewhere on your computer.
2. Deploy to Cloud Functions using the command below.

```bash
gcloud functions deploy {FUNCTION_NAME} \
    --region {REGION} \
    --allow-unauthenticated \
    --entry-point handler \
    --runtime python39 \
    --service-account {SERVICE_ACCOUNT} \
    --project {PROJECT_NAME} \
    --trigger-http \
    --env-vars-file ".env.json"
```
3. Enter the url(`{SERVER_ENDPOINT}/slack/install`) in the search bar to install Slackbot.

(I think it might be helpful) service-acoount has permisssions below:

- Cloud Functions Developer
- Cloud Functions Invoker
- Cloud SQL Client

An easier reproduction would be to insert the code below into this [tutorial](https://github.com/slackapi/bolt-python/blob/main/examples/google_cloud_functions/main.py).

```python3
flask_app = Flask(__name__)

[...]

PR #649, opened 2022-05-16
Add Google Cloud Functions adapter (ref #646)

seratch:
This pull request adds a new adapter for Google Cloud Functions users. You can use Flask adapter for it as long as you don't have the OAuth flow in your app. However, as we see at #646,  implementing the OAuth flow is far different and it would be confusing for developers. This pull request adds a new adapter that is quite similar to the AWS Lambda adapter. 

One thing developers should be aware of would be the lazy listeners are not yet available for Google Cloud Functions, unlike AWS Lambda. Thus, I've added Noop runner, which just raises an exception when a developer tries to use lazy listeners in Cloud Functions app. Also, this is an improvement from the Flask adapter way.

* [ ] Others

### DIFF
--- a/examples/google_cloud_functions/oauth_main.py
+++ b/examples/google_cloud_functions/oauth_main.py
@@ -80,7 +80,7 @@ if __name__ == "__main__":
 
 
 # Step1: Create a new Slack App: https://api.slack.com/apps
-# Bot Token Scopes: chat:write, commands, app_mentions:read
+# Bot Token Scopes: app_mentions:read,chat:write,commands
 
 # Step2: Set env variables
 # cp .env.yaml.oauth-sample .env.yaml
--- a/examples/google_cloud_functions/oauth_main.py
+++ b/examples/google_cloud_functions/oauth_main.py
@@ -1,0 +1,101 @@
+# https://cloud.google.com/functions/docs/first-python
+
+import logging
+
+from slack_bolt.oauth.oauth_settings import OAuthSettings
+
+logging.basicConfig(level=logging.DEBUG)
+
+from slack_bolt import App
+from datastore import GoogleDatastoreInstallationStore, GoogleDatastoreOAuthStateStore
+
+from google.cloud import datastore
+
+datastore_client = datastore.Client()
+logger = logging.getLogger(__name__)
+
+# process_before_response must be True when running on FaaS
+app = App(
+    process_before_response=True,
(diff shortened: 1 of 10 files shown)


seratch:
Ah, yeah I should add some tests. Will do after finishing other tasks.

slack_bolt/adapter/google_cloud_functions/handler.py
```diff
@@ -0,0 +1,64 @@
...
+    def __init__(self, app: App):  # type: ignore
+        self.app = app
+        # Note that lazy listener is not supported
+        self.app.listener_runner.lazy_listener_runner = NoopLazyListenerRunner()
+        if self.app.oauth_flow is not None:
+            self.app.oauth_flow.settings.redirect_uri_page_renderer.install_path = "?"
+
+    def handle(self, req: Request) -> Response:
+        if req.method == "GET":
+            if self.app.oauth_flow is not None:
```
stasfilin, line 51: We can remove this extra **if**

Just change 
```python
        if req.method == "GET":
            if self.app.oauth_flow is not None:
```

to 

```python
        if req.method == "GET" and self.app.oauth_flow is not None
```

seratch: @stasfilin Thanks for the suggestion!

seratch:
Updated the code (thanks for the suggestion, @stasfilin!) and added unit tests ✅

examples/google_cloud_functions/oauth_main.py
```diff
@@ -0,0 +1,101 @@
...
+
+    @flask_app.route("/function", methods=["GET", "POST"])
+    def handle_anything():
+        return handler.handle(request)
+
+    flask_app.run(port=3000)
+
+
+# Step1: Create a new Slack App: https://api.slack.com/apps
+# Bot Token Scopes: chat:write, commands, app_mentions:read
```
filmaj, line 83: FYI the .env.yaml.oauth-sample has a different list than this; it includes `channels.history`.

filmaj (approved):
Code LGTM but I did not create a test app using this as salesforce does not allow me to create Google Cloud projects 😑

examples/google_cloud_functions/oauth_main.py
seratch: Thanks! I've updated the part.

PR #679, opened 2022-07-11
Allow lazy function invocation to target version/alias

angrychimp:
Currently lazy responses are only targeting the `$LATEST` version of a given AWS function. This prevents meaningful A/B testing if you're using function versions/aliases to compare code changes. Adjusting the `invoke` method to use the function ARN from the current execution allows downstream code parity.

This change adds a variable (`invoked_function_arn`) to the request context, then uses that value when invoking the sub-request.

### Category

* [X] Adapters in `slack_bolt.adapter`
* [ ] Others

### DIFF
--- a/slack_bolt/adapter/aws_lambda/lazy_listener_runner.py
+++ b/slack_bolt/adapter/aws_lambda/lazy_listener_runner.py
@@ -23,7 +23,7 @@ class LambdaLazyListenerRunner(LazyListenerRunner):
         headers["x-slack-bolt-lazy-function-name"] = request.lazy_function_name  # not an array
         event["method"] = "NONE"
         invocation = self.lambda_client.invoke(
-            FunctionName=request.context["aws_lambda_function_name"],
+            FunctionName=request.context["aws_lambda_invoked_function_arn"],
             InvocationType="Event",
             Payload=json.dumps(event),
         )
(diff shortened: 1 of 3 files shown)


slack_bolt/adapter/aws_lambda/lazy_listener_runner.py
```diff
@@ -23,7 +23,7 @@ def start(self, function: Callable[..., None], request: BoltRequest) -> None:
         headers["x-slack-bolt-lazy-function-name"] = request.lazy_function_name  # not an array
         event["method"] = "NONE"
         invocation = self.lambda_client.invoke(
-            FunctionName=request.context["aws_lambda_function_name"],
+            FunctionName=request.context["aws_lambda_invoked_function_arn"],
```
seratch, line 26: This should work. The lambda invocation API accepts any of function name, function ARN, and partial ARN. https://docs.aws.amazon.com/cli/latest/reference/lambda/invoke.html#options

seratch (approved):
Thanks for improving this! LGTM

seratch:
@angrychimp The change looks great to me already but this change requires a small update in the unit test code. Could you update the test code to be consistent with your code change?

angrychimp:
Not sure how I missed that, but thanks for pointing it out. I've added the updated test script to the PR. I'm getting a seg-fault running the "run_tests" script which I believe is just an issue with my environment, but while I figure that out let me know if it looks like there's anything else I'm missing.

> @angrychimp The change looks great to me already but this change requires a small update in the unit test code. Could you update the test code to be consistent with your code change?

PR #696, opened 2022-08-05
Future

WilliamBergamin:
Disclaimer, my bad this is a huge PR and the CLI and remote function stuff got mangled together.

The goal of this PR is to get feedback and add basic remote function support to bolt python. This contribution only includes support for synchronous architecture and does not include the related documentation.

This also include very basic support for the slackcli along with its tests.

[Here are the modifications to the python template](https://github.com/slack-samples/bolt-python-starter-template/tree/future/listeners/functions) I've done to use these changes

#### cli stuff
- [slack_bolt/cli](https://github.com/slackapi/bolt-python/tree/future/slack_bolt/cli)
- [tests/slack_bolt/cli](https://github.com/WilliamBergamin/bolt-python/tree/future/tests/slack_bolt/cli)

#### function stuff
- Every file affected by changes excluding [cli stuff](#cli-stuff)

### Testing
Test these changes manually
#### Setup
you will need a file structure that resembles:
```
.
├── bolt-python-starter-future         # https://github.com/slack-samples/bolt-python-starter-template/tree/future
└── bolt-python-future-fork            # https://github.com/WilliamBergamin/bolt-python/tree/future
```
helpful commands
``` zsh
git clone https://github.com/slack-samples/bolt-python-starter-template/tree/future
git clone https://github.com/WilliamBergamin/bolt-python/tree/future
```

##### Bolt
1. Navigate to your `bolt-python-future-fork` folder
2. set up a python env
    ```zsh
    brew install pyenv
    pyenv install 3.9.6
    python -m venv env_3.9.6
    source env_3.9.6/bin/activate
    ```
3. build the package using the script
    ```zsh
     bash scripts/build_pypi_package.sh 
     ```
##### application
1. Navigate to your `bolt-python-starter-future` folder
2. Set up a pyenv follow the same steps as [Bolt step 2](#bolt)
3. ensure that this dependency is present in the `requirements.txt` file
    ```zsh
     ../bolt-python-future-fork/dist/slack_bolt-1.14.3-py2.py3-none-any.whl
     ```
     this tells pip where your local build of python-bolt is located
4. install the dependencies
    ```zsh
     ```

#### Run
1. With slack cli (else you need to install the app traditionally)
    ```zsh
     slack install
     slack run
     ```
2. Add the app _**Bolt Template App (dev)**_ to a channel in your workspace
3. Find and trigger the Reverse shortcut to start the function
Screen Shot 2022-08-08 at 3 25 01 PM

### Feed back
I would like to figure out if these changes yield the proper pythonic user experience.
example: Is passing `complete_error` and `complete_success` a correct python way of resolving a function 

* [ ] Others


slack_bolt/context/context.py
```diff
@@ -7,8 +7,17 @@
 from slack_bolt.context.base_context import BaseContext
 from slack_bolt.context.respond import Respond
 from slack_bolt.context.say import Say
+from slack_bolt.context.complete_success import CompleteSuccess
+from slack_bolt.context.complete_error import CompleteError
 from slack_bolt.util.utils import create_copy
 
+CLIENT: str = "client"
```
WilliamBergamin, line 14: I felt like these values were used many times in the file, to prevent typos I consolidated them in variables

slack_bolt/logger/messages.py
```diff
@@ -186,33 +197,14 @@ def warning_unhandled_request(  # type: ignore
             or filtered_body.get("view", {}).get("callback_id")  # type: ignore
             or "your-callback-id"
         )
-        return _build_unhandled_request_suggestion(
-            default_message,
-            f"""
```
WilliamBergamin, line 191: The sample code snippets felt like they were cluttering the function, I moved them to a separate file

slack_bolt/cli/error/__init__.py
```diff
@@ -0,0 +1,2 @@
+class CliError(Exception):
```
seratch, line 1: Can move this class to https://github.com/slackapi/bolt-python/blob/main/slack_bolt/error/__init__.py ?

tests/slack_bolt/cli/test_get_manifest.py
```diff
@@ -0,0 +1,24 @@
+import json
+import pytest
+
+from slack_bolt.cli.get_manifest import get_manifest
+
+
+class TestGetManifest:
+    def test_get_manifest(self):
+        working_directory = "tests/slack_bolt/cli/test_directory"
```
seratch, line 9: test_directory sounds like a place where we may create and delete files as a sandbox. If I understand correctly, the files under the directory looks like example apps used for the testing. How about renaming to sample/example_app or test_app?

seratch (changes requested):
Thanks for working on this. You don't need to add asyncio support at this moment but when we ship this feature, we definitely must have the support too.

srajiang (commented):
@WilliamBergamin 
Hey there! Excited to dig into this, but have some asks for you. Can you please: 
* Add a Testing section with some steps for manual testing the functionality
* Provide an overview (just because it's a large PR on what files contain what changes?

WilliamBergamin:
Thanks for the feedback @srajiang, Ive added this information at the top, let me know if you have any issues setting up project locally or want more info 😄

slack_bolt/cli/error/__init__.py
WilliamBergamin: I've added this change, thanks

tests/slack_bolt/cli/test_get_manifest.py
WilliamBergamin: I've renamed the folder to test_app, thanks

slack_bolt/context/complete_success/complete_success.py
```diff
@@ -0,0 +1,35 @@
...
+        outputs: Union[str, dict],
+    ) -> SlackResponse:
+        if self._can_complete():
+            if isinstance(outputs, dict) or isinstance(outputs, str):
+                # TODO add this new api call to the sdk and use it here
+                return self.client.api_call(
+                    "functions.completeSuccess",
+                    json={"outputs": outputs, "function_execution_id": self.function_execution_id},
+                )
+            raise ValueError(f"The outputs arg is unexpected type ({type(outputs)})")
```
filmaj, line 30: Same feedback as previous comment: let's state the required type, and the context for the error.

slack_bolt/context/context.py
```diff
@@ -33,94 +42,132 @@ def to_copyable(self) -> "BoltContext":
 
     @property
     def client(self) -> Optional[WebClient]:
-        """The `WebClient` instance available for this request.
+        f"""The `WebClient` instance available for this request.
 
-            @app.event("app_mention")
+            @app.{CLIENT}("app_mention")
```
filmaj, line 47: I think this change might be a mistake?

filmaj (approved):
Hey nice work with this! I followed the testing steps and it worked great. One thing I did notice, not sure why this happens, but,  after I ran `slack run` into my production test workspace, I got a DM from Slackbot in that workspace:

> Your workspace has been using /sample-command to kick off certain actions with Bolt Template App. Recently Bolt Template App (dev) was installed to Filbox with the same command. Now when people enter /sample-command, it'll run the action set up by Bolt Template App (dev). You can learn more about [managing app installation settings](https://slack.com/help/articles/201259356-Slash-commands-in-Slack) in the Help Center.

Never seen that before but figured I'd mention it.

Re: "is this the pythonic way of doing this" I will defer to @seratch on that question 😅 . I used to work in Python a lot at one company several years ago, and one thing I learned from that crew is that there is no one way of doing things in Python 😄 I think the CLI hook implementation using python scripts is fine. Callbacks as a response mechanism also seems OK to me?

I left a few minor comments as well.

srajiang (commented):
Hey @WilliamBergamin - this is looking great! 

Some edge case behaviors I wanted to raise with you: So long as they don't violate some best practices in Python, we can aim for consistency in the hook behavior. 

* get-manifest supports finding a manifest file outside of the project root. This is something I put in place for JS / TS because of the variety of ways you might set up your project folder (including in a `src` folder). Curious what your thoughts whether support for this is needed in Python.

* start supports customization of app entry-point path at the project level. I noticed you are honoring a potential SLACK_CLI_CUSTOM_FILE_PATH passed down through env (to date, CLI doesn't send this and I'm kind of on the side of the fence of removing this support and letting the dev set this at the project level). I'm curious is that something Python devs would expect to be able to do? In node, for example, we let them set their entrypoint via the package.json and the hook honors this entrypoint.

slack_bolt/context/context.py
WilliamBergamin: I think `@app.{CLIENT}("app_mention")` should render to `@app.client("app_mention")` but I could be wrong, is this what you think the mistake could be?

WilliamBergamin:
Thanks for dropping in some of your wisdom @srajiang !!

> get-manifest supports finding a manifest file outside of the project root. This is something I put in place for JS / TS because of the variety of ways you might set up your project folder (including in a src folder). Curious what your thoughts whether support for this is needed in Python.

This feature can be helpful for the user 💯  it is definitely something we can implement in python I just want to understand the scope better, should the script accept a `folder path` in an env variable (like SLACK_CLI_CUSTOM_FILE_PATH) or as a parameter? and when you mention _including in a `src` folder_ this means searching through the subdirectory to find the file?

> start supports customization of app entry-point path at the project level. I noticed you are honoring a potential SLACK_CLI_CUSTOM_FILE_PATH passed down through env (to date, CLI doesn't send this and I'm kind of on the side of the fence of removing this support and letting the dev set this at the project level). I'm curious is that something Python devs would expect to be able to do? In node, for example, we let them set their entrypoint via the package.json and the hook honors this entrypoint.

Maybe @seratch can peer review me on this, from my understanding pip is a simple dependency manager, unlike npm, gradle or maven, pip or python do not have knowledge surrounding the project. There isn't an equivalent to `package.json` file builtin to python. Thus the user needs to somehow provide the entrypoint file to execute the process, using env variable like SLACK_CLI_CUSTOM_FILE_PATH is a simple and scalable way to pass values to the script. But we can use other approaches.

slack_bolt/context/complete_error/complete_error.py
```diff
@@ -0,0 +1,34 @@
...
+        self,
+        message: str,
+    ) -> SlackResponse:
+        if self._can_complete():
+            if isinstance(message, str):
+                # TODO add this new api call to the sdk and use it here
+                return self.client.api_call(
+                    "functions.completeError", json={"error": message, "function_execution_id": self.function_execution_id}
+                )
+            raise ValueError(f"The message arg is unexpected type ({type(message)})")
```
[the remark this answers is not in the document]
WilliamBergamin: Ive implemented this let me know if this is what you had in mind

slack_bolt/context/complete_success/complete_success.py
WilliamBergamin: I've implemented this let me know if this is what you had in mind

slack_bolt/kwargs_injection/args.py
```diff
@@ -69,6 +71,10 @@ def handle_buttons(ack, respond, logger, context, body, client):
     """`say()` utility function, which calls `chat.postMessage` API with the associated channel ID"""
     respond: Respond
     """`respond()` utility function, which utilizes the associated `response_url`"""
+    complete_success: CompleteSuccess
+    """`complete_success()` utility function, which utilizes the associated `function_execution_id`"""
+    complete_error: CompleteError
+    """`complete_error()` utility function, which utilizes the associated `function_execution_id`"""
```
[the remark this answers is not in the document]
WilliamBergamin: I've implemented this let me know if this might be to long

slack_bolt/context/context.py
srajiang: @WilliamBergamin -Should it resolve to "event" though and not to "client"?

WilliamBergamin: Ohh wow, completely missed that thanks

srajiang:
> This feature can be helpful for the user 💯 it is definitely something we can implement in python, I just want to understand the scope better, should the script accept a folder path in an env variable (like SLACK_CLI_CUSTOM_FILE_PATH) or as a parameter? and when you mention including in a src folder this means searching through the subdirectory to find the file?

There are different options for doing this that aren't mutually exclusive. 

One way is to require developers to supply the manifest path as environment variables (in that case, perhaps it doesn't make sense to use SLACK_CLI_CUSTOM_FILE_PATH, since SLACK_CLI_* preface indicates it's being passed along from the CLI context). You'd want to name it something different I think. 

AND / OR you could make the hook do a little more by doing a search through the project and looking for the manifest. That's what I've opted for with Bolt JS, mostly just to avoid giving developers another hurdle in setup. The thinking here is that having a manifest file in your project in the first place is a CLI / 2.0 requirement. If you only had the env variable supported then it'd become a requirement for the dev to set that up first thing. I think that's less of a good DX for a first time user. 

Re: entrypoint path in the env That's something I do think makes sense for devs to want to configure or setup based on other needs.

seratch:
@srajiang 
As @WilliamBergamin mentioned, pip is just a dependency resolver. This does not manage anything about your Python project. Also, even when you manage your Python project using more sophisticated tools such as Poetry, those tools are only interested in which Python runtime is used and where to load source files.

As long as the CLI-related paths never affect where to place source files and which runtime to use, you can use any paths for CLI regardless of how to load the paths (env variables or whatever).

seratch (approved):

srajiang:
> source files and which runtime to use, you can use any paths for CLI regardless of how to load the paths

We shouldn't be using SLACK_CLI_* preface for variables that the developer is expected to set, since that is the prefix the CLI uses when it sets environment variables.

WilliamBergamin:
@srajiang Ive modified the `get_manifest` hook to search in subfolders of the working directory for the `manifest.json`, One important thing to address is that I am attempting to ignore certain Folders (_hidden folders, generated folders, dependencies, tests_)
[This is the new file in question](https://github.com/WilliamBergamin/bolt-python/blob/future/slack_bolt/cli/get_manifest.py) let me know what you think of this solution


As for the SLACK_CLI_* env variables I was under the impression these were set by the cli before executing the hook, I will take this off line to get more info.

PR #715, opened 2022-08-31
Add basic remote function interactivity

WilliamBergamin:
The goal of this PR is to add remote function interactivity support to bolt python. This only includes basic action interactivity. This contribution does not include the related documentation.

[Here are the modifications to the python template](https://github.com/slack-samples/bolt-python-starter-template/tree/future) that allow to test the interactivity with a simple approve/deny action

### Testing
Test these changes manually
#### Setup
you will need a file structure that resembles:
```
.
├── bolt-python-starter-future        # https://github.com/slack-samples/bolt-python-starter-template/tree/future
└── bolt-python-future-fork            # https://github.com/WilliamBergamin/bolt-python/tree/future
```
helpful commands
``` zsh
git clone https://github.com/slack-samples/bolt-python-starter-template/tree/future-async
git clone https://github.com/WilliamBergamin/bolt-python/tree/future
```

##### Bolt
1. Navigate to your `bolt-python-future-fork` folder
2. set up a python env
    ```zsh
    brew install pyenv
    pyenv install 3.9.6
    python -m venv env_3.9.6
    source env_3.9.6/bin/activate
    ```
3. build the package using the script
    ```zsh
     bash scripts/build_pypi_package.sh 
     ```
##### application
1. Navigate to your `bolt-python-starter-future` folder
2. Set up a pyenv follow the same steps as [Bolt step 2](#bolt)
3. ensure that this dependency is present in the `requirements.txt` file
    ```zsh
     ../bolt-python-future-fork/dist/slack_bolt-1.14.3-py2.py3-none-any.whl
     ```
     this tells pip where your local build of python-bolt is located
4. install the dependencies
    ```zsh
     ```

#### Run
1. With slack cli (else you need to install the app traditionally)
    ```zsh
     slack trigger create --trigger-def "./trigger/approve_me.json"
     slack run
     ```
2. Add the app _**Bolt Template App (dev)**_ to a channel in your workspace
3. Send the _trigger link_ in the channel and execute it with the `run` command 

### Usage

```python
from slack_sdk import WebClient
from slack_bolt import Complete, Ack
from logging import Logger

from app import app


APPROVE_ID = "approve_action_id"
DENY_ID = "deny_action_id"

@app.function("approve_me")
def approve_me(event, client: WebClient, complete: Complete, logger: Logger):
    try:
        channel = event["inputs"]["channel"]

        client.chat_postMessage(
            channel=channel,
            text='Approve me please',
            blocks=[
                {
                    "type": "section",
                    "text": {
                        "type": "mrkdwn",
                    	"text": "*Approve me please* :wink:"
                    }
                },
                {
                    "type": 'actions',
                    "block_id": 'approve-deny-buttons',
                    "elements": [
                        {
                            "type": 'button',
                            "text": {
                                "type": 'plain_text',
                                "text": 'Approve',
                            },
                            "action_id": APPROVE_ID,
                            "style": 'primary',
                        },
                        {
                            "type": 'button',
                            "text": {
                                "type": 'plain_text',
                                "text": 'Deny',
                            },
                            "action_id": DENY_ID,
                            "style": 'danger',
                        }
                    ]
                }
            ])
    except Exception as e:
        logger.error(e)
        complete(error="Cannot request approval")
        raise e


@approve_me.action(APPROVE_ID)
def approve_action(ack: Ack, client: WebClient, body, complete: Complete, logger: Logger):
    try:
        ack()
        blocks = body["message"]["blocks"][:-1]
        blocks.append({
            "type": 'context',
            "elements": [
                {
                    "type": 'mrkdwn',
                    "text": f":white_check_mark: I have been approved",
                },
            ],
        })

        client.chat_update(
            channel=body["container"]["channel_id"],
            ts=body["container"]["message_ts"],
            text="I have been approved",
            blocks=blocks
        )

        complete()
    except Exception as e:
        logger.error(e)
        complete("Cannot request approval")
        raise e


@approve_me.action(DENY_ID)
def deny_action(ack: Ack, client: WebClient, body, complete: Complete, logger: Logger):
    try:
        ack()
        blocks = body["message"]["blocks"][:-1]
        blocks.append({
            "type": 'context',
            "elements": [
                {
                    "type": 'mrkdwn',
                    "text": f":no_entry: I have been denied",
                },
            ],
        })

        client.chat_update(
            channel=body["container"]["channel_id"],
            ts=body["container"]["message_ts"],
            text="I have been denied",
            blocks=blocks
        )

        complete()
    except Exception as e:
        logger.error(e)
        complete("Cannot request approval")
        raise e

# or 

approve_me_function: Function = app.function("approve_me")(approve_me)
approve_me_function.action(APPROVE_ID)(approve_action)
approve_me_function.action(DENY_ID)(deny_action)

```

### Feedback
Feedback on naming and the folder structure would be greatly appreciated

* [ ] Others


slack_bolt/context/base_context.py
```diff
@@ -91,6 +91,11 @@ def authorize_result(self) -> Optional[AuthorizeResult]:
         """The authorize result resolved for this request."""
         return self.get("authorize_result")
 
+    @property
+    def bot_access_token(self) -> Optional[str]:
```
WilliamBergamin, line 95: Decided to keep this because the original bot-token is still accessible from the function but hidden

seratch: @WilliamBergamin can you make the name consistent with bolt-js? https://github.com/slackapi/bolt-js/pull/1567#discussion_r959075081

slack_bolt/context/complete/async_complete.py
```diff
@@ -0,0 +1,62 @@
...
+        if isinstance(error, str):
+            # TODO add this new api call to the sdk and use it here
+            return await self.client.api_call(
+                "functions.completeError", json={"error": error, "function_execution_id": self.function_execution_id}
+            )
+
+        # TODO add this new api call to the sdk and use it here
+        return await self.client.api_call(
+            "functions.completeSuccess",
+            json={"outputs": {}, "function_execution_id": self.function_execution_id},
```
seratch, line 54: Use the outputs value, not the empty dict

slack_bolt/listener_matcher/builtins.py
```diff
@@ -1,543 +0,0 @@
-# pytype: skip-file
```
seratch, line 1: Making these refactoring as part of future branch will make merging it into main more difficult. These big changes are not necessary for implementing functions. Can you do this separately towards main branch?

seratch (changes requested):

slack_bolt/context/complete/async_complete.py
WilliamBergamin: Good catch @seratch 💯

slack_bolt/context/complete/internals.py
```diff
@@ -3,3 +3,12 @@
 
 def can_not_complete(self: Any) -> bool:
     return not hasattr(self, "client") or self.client is None or self.function_execution_id is None
+
+
+def get_kwargs_msg(self: Any) -> str:
```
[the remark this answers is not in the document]
WilliamBergamin: Thanks, I felt like there was room for better naming here

slack_bolt/listener_matcher/builtins.py
WilliamBergamin: That makes sense, I've removed these changes ✂️ 

It should also make the PR easier to read

slack_bolt/app/app.py
```diff
@@ -1247,32 +1247,16 @@ def enable_token_revocation_listeners(self) -> None:
...
-    def _to_listener_functions(
-        kwargs: dict,
-    ) -> Optional[Sequence[Callable[..., Optional[BoltResponse]]]]:
-        if kwargs:
-            functions = [kwargs["ack"]]
-            for sub in kwargs["lazy"]:
-                functions.append(sub)
-            return functions
-        return None
+        client_per_request: WebClient = WebClient(
```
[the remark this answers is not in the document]
WilliamBergamin: Yess rebased the future branch to have this changes, thanks this should also clean up the PR 💯

slack_bolt/app/app.py
```diff
@@ -794,7 +792,7 @@ def __call__(*args, **kwargs):
 
     def function(
         self,
-        callback_id: Union[str, Pattern],
+        callback_id: Union[str, Function],
```
seratch, line 795: Yeah, I agree that we don't need to support regexp here

slack_bolt/middleware/function_token/function_listener_token.py
```diff
@@ -0,0 +1,24 @@
...
+        # As this method is not supposed to be invoked by bolt-python users,
+        # the naming conflict with the built-in one affects
+        # only the internals of this method
+        next: Callable[[], BoltResponse],
+    ) -> BoltResponse:
+        if req.context.bot_access_token:
+            req.context.client.token = req.context.bot_access_token
+            return next()
+
+        # As the token was not found for this function call, skip running the listener
```
seratch, line 23: Is it safe to assume that bot_access_token always exists for function events? If not, we still should call `next()` in any case here.

slack_bolt/middleware/function_token/function_listener_token.py
```diff
@@ -0,0 +1,24 @@
+from typing import Callable
+
+from slack_bolt.request import BoltRequest
+from slack_bolt.response import BoltResponse
+from slack_bolt.middleware.middleware import Middleware
+
+
+class FunctionToken(Middleware):  # type: ignore
```
seratch, line 8: How about making the name a bit clearer?
```suggestion
class AttachingFunctionToken(Middleware):  # type: ignore
```

slack_bolt/context/base_context.py
seratch: The prop name is not yet updated -- we should use slack_function_bot_access_token instead.

slack_bolt/context/complete/complete.py
```diff
@@ -0,0 +1,47 @@
...
+from slack_bolt.context.complete.internals import can_not_complete, get_kwargs_for_logging, get_name_for_logging
+
+
+class Complete:
+    client: Optional[WebClient]
+    function_execution_id: Optional[str]
+
+    def __init__(
+        self,
+        client: Optional[WebClient],
```
seratch, line 14: Can the `client` argument be None? If not, we can change the type to `WebClient`

slack_bolt/request/internals.py
```diff
@@ -132,6 +132,18 @@ def extract_function_execution_id(payload: Dict[str, Any]) -> Optional[str]:
...
     if payload.get("event") is not None:
         return extract_function_execution_id(payload["event"])
+    if payload.get("function_data") is not None:
+        return extract_function_execution_id(payload["function_data"])
+    if payload.get("execution_id") is not None:
+        return payload.get("execution_id")
+    return None
+
+
+def extract_bot_access_token(payload: Dict[str, Any]) -> Optional[str]:
```
seratch, line 142:
```suggestion
def extract_slack_function_bot_access_token(payload: Dict[str, Any]) -> Optional[str]:
```

slack_bolt/request/internals.py
```diff
@@ -152,6 +164,9 @@ def build_context(context: BoltContext, body: Dict[str, Any]) -> BoltContext:
     function_execution_id = extract_function_execution_id(body)
     if function_execution_id:
         context["function_execution_id"] = function_execution_id
+    bot_access_token = extract_bot_access_token(body)
+    if bot_access_token:
+        context["bot_access_token"] = bot_access_token
```
seratch, line 169: slack_function_bot_access_token

slack_bolt/request/payload_utils.py
```diff
@@ -182,6 +182,17 @@ def to_step(body: Dict[str, Any]) -> Optional[Dict[str, Any]]:
...
 
+# -------------------
+# Function
+# -------------------
+
+
+def is_function_interactivity(body: Dict[str, Any]) -> bool:
+    if "function_data" in body:
+        return True
+    return False
```
seratch, line 193:
```suggestion
    return "function_data" in body
```

slack_bolt/util/utils.py
```diff
@@ -88,3 +89,14 @@ def get_name_for_callable(func: Callable) -> str:
 
 def get_arg_names_of_callable(func: Callable) -> List[str]:
     return inspect.getfullargspec(inspect.unwrap(func)).args
+
+
+def to_listener_functions(
```
seratch, line 94: Can you add some comments to describe what this function is?

tests/mock_web_api_server.py
```diff
@@ -28,6 +29,9 @@ def is_valid_token(self):
     def is_valid_user_token(self):
         return "Authorization" in self.headers and str(self.headers["Authorization"]).startswith("Bearer xoxp-")
 
+    def is_valid_bot_access_token(self):
```
seratch, line 32:
```suggestion
    def is_valid_slack_function_bot_access_token(self):
```

slack_bolt/function/function.py
```diff
@@ -0,0 +1,86 @@
...
+        self,
+        constraints: Union[str, Pattern, Dict[str, Union[str, Pattern]]],
+        matchers: Optional[Sequence[Callable[..., bool]]] = None,
+        middleware: Optional[Sequence[Union[Callable, Middleware]]] = None,
+    ) -> Callable[..., Optional[Callable[..., Optional[BoltResponse]]]]:
+        """Registers a new action listener to your function. This method can be used as either a decorator or a method.
+
+            @app.function("reverse")
+            def reverse_string(event, complete_success: CompleteSuccess, complete_error: CompleteError):
+                complete_error("There is no error")
```
seratch, line 52:
```suggestion
            @app.function("request-approval")
            def handle_request_approvals(event, complete: Complete):
                complete(outputs={})
```

slack_bolt/function/function.py
```diff
@@ -0,0 +1,86 @@
...
+
+            # Use this method as a decorator
+            @reverse_string.action("approve_button")
```
seratch, line 55:
```suggestion
            # Use this method as a decorator
            request_apprval = app.function("request-approval")
            @request_apprval
            def handle_request_apprval_events():
                pass
               
            @request_apprval.action("approve_button")
```

slack_bolt/function/function.py
```diff
@@ -0,0 +1,86 @@
...
+            def update_message(ack):
+                ack()
+
+            # Pass a function to this method
+            reverse_string.action("approve_button")(update_message)
```
seratch, line 60: same as above

slack_bolt/function/function.py
```diff
@@ -0,0 +1,86 @@
...
+from slack_bolt.listener_matcher import builtins as builtin_matchers
+
+from slack_bolt.response import BoltResponse
+from slack_bolt.middleware import Middleware
+from slack_bolt.middleware.function_token import FunctionToken
+
+from slack_bolt.util.utils import to_listener_functions
+
+
+class Function:
```
seratch, line 13: If we don't support the view method for now, can you add TODO comment about it?

slack_bolt/app/app.py
WilliamBergamin: Yess but I think Function might be a typo, Ill look into this

slack_bolt/middleware/function_token/function_listener_token.py
WilliamBergamin: I think the only situation where `bot_access_token` is not present is when this middleware is used on a _none function related request_ in this case it could just handle the listener normally, Ill add this change ☄️

slack_bolt/middleware/function_token/function_listener_token.py
WilliamBergamin: Love this name, was struggling to find something meaningful 🥇

slack_bolt/context/complete/complete.py
WilliamBergamin: Good catch, here the client should never be `None` if it is an error should be thrown

slack_bolt/context/base_context.py
WilliamBergamin: @seratch @srajiang I made the change to `slack_function_bot_access_token` but It feels long, I was thinking of changing it to `function_bot_access_token` but it would not be consistent with bolt-js

slack_bolt/function/function.py
WilliamBergamin: I was going to add this in a different PR, But will add this as a TODO

slack_bolt/util/utils.py
WilliamBergamin: Renamed the function to `get_callables_from_kwargs` instead of adding a comment

slack_bolt/context/base_context.py
srajiang: @WilliamBergamin I know, it does feel long. I guess if you need a decider reasoning, I'd opt for consistency and clarity over brevity in this one case - in future when need to propagate changes across SDKs, it'll make finding things a little easier. SlackFunctionBotAccessToken in JS and slack_function_bot_access_token in Python are very unmistakeable.

srajiang: Ah, but on looking at your `Function` class, I see you've already opted to go with just `Function` over `SlackFunction`, so I do see how you could opt to be internally consistent here and omit `slack`.

WilliamBergamin: @srajiang I was also considering renaming `Function` to `SlackFunction` and `app.function` to `app.slack_function` but it felt long once I made the changes, now is a good time to make a more final decision on what should be implemented

WilliamBergamin: I rename to `slack_function` `function` can really become ambiguous and this is how it is done in bolt-js

srajiang: So - `app.function('id', new SlackFunction( ...))` is what we support at the top level, for brevity reasons only if that's helpful. We went with `SlackFunction` for the class because `new Function` just seemed wrong from JS perspective (everything is a function in this language + there's a reserved word so that makes lookup possibly a little more annoying for devs).

`SlackFunction` is what Deno Slack developers will instantiate as well for their handlers, but again we're all in the same JS boat together so that's probably why we've ended up going with the same naming here. 

I think at the end of the day, if `Function` can be considered by a Python dev to be clear and intuitive as to what they're creating, then we're okay. And of course we will always have to have our own versions of docs, so I defer to you on both where to strike the balance with this!

WilliamBergamin: Ahh I see, naming the class `SlackFunction` and using `app.function` should result in a clean and intuitive python user experience 🥇

slack_bolt/listener_matcher/builtins.py
```diff
@@ -188,18 +205,58 @@ def func(body: Dict[str, Any]) -> bool:
...
+
+        def func(body: Dict[str, Any]) -> bool:
+            return _function_block_action(callback_id, constraints, body)
+
+        return build_listener_matcher(func, asyncio, base_logger)
+
+    elif "type" in constraints:
+        action_type = constraints["type"]
+        if action_type == "block_actions":
+            return function_block_action(callback_id, constraints, asyncio)
```
filmaj, line 228: should asyncio be passed in here? same for the next conditional down

slack_bolt/request/payload_utils.py
```diff
@@ -182,6 +182,15 @@ def to_step(body: Dict[str, Any]) -> Optional[Dict[str, Any]]:
     return None
 
 
+# -------------------
+# Function
+# -------------------
+
+
+def is_function_interactivity(body: Dict[str, Any]) -> bool:
+    return "function_data" in body
```
filmaj, line 191: Technically the presence of `function_data` in the payload body can signify _either_ interactivity events (block_actions, view_submission, view_closed) but it could ALSO signal the function-executed event. Not sure if that's relevant here.

filmaj (approved):
Worked well for me! Looks good. Left a couple of very minor comments.

slack_bolt/listener_matcher/builtins.py
WilliamBergamin: Good catch 💯  It is not used in this PR but it should be used in a PR coming soon, if it is not it will be removed

slack_bolt/context/async_context.py
```diff
@@ -133,41 +131,20 @@ async def handle_button_clicks({ACK}, {RESPOND}):
         return self[RESPOND]
 
     @property
-    def complete_success(self) -> AsyncCompleteSuccess:
-        f"""`{COMPLETE_SUCCESS}()` function for this request.
+    def complete(self) -> AsyncComplete:
+        f"""`{COMPLETE}()` function for this request.
```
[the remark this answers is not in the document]
WilliamBergamin: Ive added this and added additional information about this function

slack_bolt/context/context.py
```diff
@@ -135,39 +133,20 @@ def handle_button_clicks(ack, {RESPOND}):
         return self[RESPOND]
 
     @property
-    def complete_success(self) -> CompleteSuccess:
-        f"""`{COMPLETE_SUCCESS}()` function for this request.
+    def complete(self) -> Complete:
+        f"""`{COMPLETE}()` function for this request.
```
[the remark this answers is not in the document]
WilliamBergamin: Ive added this and added additional information about this function

slack_bolt/request/payload_utils.py
WilliamBergamin: Good catch will add more validations here seems like bolt validates the type as well 🥇

slack_bolt/util/utils.py
```diff
@@ -88,3 +89,14 @@ def get_name_for_callable(func: Callable) -> str:
 
 def get_arg_names_of_callable(func: Callable) -> List[str]:
     return inspect.getfullargspec(inspect.unwrap(func)).args
+
+
+def extract_callables(
```
seratch, line 94: I think that this new name is too general. The list of callable that this method returns is only useable as listener functions. How about naming to `extract_listener_callables` or `extract_listener_functions`?

slack_bolt/context/async_context.py
```diff
@@ -133,41 +131,31 @@ async def handle_button_clicks({ACK}, {RESPOND}):
...
 
-            @app.function("reverse")
-            async def handle_button_clicks(context):
-                await context.{COMPLETE_ERROR}("an error spawned")
-
-            @app.function("reverse")
-            async def handle_button_clicks({COMPLETE_ERROR}):
-                await {COMPLETE_ERROR}("an error spawned")
+        Kwargs:
+            ** THIS FUNCTION ONLY ACCEPTS ONE OF THE KWARGS BELOW **
```
seratch, line 152: We should check if this additional statement line works as you expect in the API documents. You can generate https://slack.dev/bolt-python/api-docs/slack_bolt/ on your local machine by running `./scripts/generate_api_docs.sh` @WilliamBergamin Can you check this?

slack_bolt/context/complete/complete.py
WilliamBergamin: On `completeSuccess` if outputs is `None` slack interprets it as an error, I added `outputs if outputs else {}` to prevent this error use case

slack_bolt/context/context.py
```diff
@@ -135,39 +133,31 @@ def handle_button_clicks(ack, {RESPOND}):
...
             @app.function("reverse")
-            def handle_button_clicks(context):
-                context.{COMPLETE_SUCCESS}({{"stringReverse":"olleh"}})
+            async def handle_button_clicks(context):
+                context.{COMPLETE}(outputs={{"stringReverse":"olleh"}})
 
             @app.function("reverse")
-            def handle_button_clicks({COMPLETE_SUCCESS}):
-                {COMPLETE_SUCCESS}({{"stringReverse":"olleh"}})
+            async def handle_button_clicks({COMPLETE}):
```
seratch, line 147: This should not be an async funtion.

slack_bolt/slack_function/slack_function.py
```diff
@@ -0,0 +1,89 @@
...
+        Args:
+            constraints: The conditions that match a request payload
+            matchers: A list of listener matcher functions.
+                Only when all the matchers return True, the listener function can be invoked.
+            middleware: A list of lister middleware functions.
+                Only when all the middleware call `next()` method, the listener function can be invoked.
+        """
+
+        middleware = list(middleware) if middleware else []
+        middleware.insert(0, AttachingFunctionToken())
```
seratch, line 76: If you have a few listeners for function events, this middleware can be executed a few times. Can you move this to the global middleware list, not appending as a listener middleware to each listener?

slack_bolt/middleware/attaching_function_token/attaching_function_token.py
```diff
@@ -0,0 +1,20 @@
...
+class AttachingFunctionToken(Middleware):  # type: ignore
+    def process(
+        self,
+        *,
+        req: BoltRequest,
+        resp: BoltResponse,
+        # This method is not supposed to be invoked by bolt-python users
+        next: Callable[[], BoltResponse],
+    ) -> BoltResponse:
+        if req.context.slack_function_bot_access_token:
```
seratch, line 17:
```suggestion
        if req.context.slack_function_bot_access_token is not None:
```

slack_bolt/request/internals.py
```diff
@@ -152,6 +164,9 @@ def build_context(context: BoltContext, body: Dict[str, Any]) -> BoltContext:
     function_execution_id = extract_function_execution_id(body)
     if function_execution_id:
         context["function_execution_id"] = function_execution_id
+    slack_function_bot_access_token = extract_slack_function_bot_access_token(body)
+    if slack_function_bot_access_token:
```
seratch, line 168:
```suggestion
    if slack_function_bot_access_token is not None:
```

slack_bolt/request/internals.py
```diff
@@ -152,6 +164,9 @@ def build_context(context: BoltContext, body: Dict[str, Any]) -> BoltContext:
     function_execution_id = extract_function_execution_id(body)
     if function_execution_id:
```
seratch, line 165:
```suggestion
    if function_execution_id is not None:
```

tests/scenario_tests/test_function_block_action.py
```diff
@@ -0,0 +1,213 @@
...
+            "content-type": ["application/json"],
+            "x-slack-signature": [self.generate_signature(body, timestamp)],
+            "x-slack-request-timestamp": [timestamp],
+        }
+
+    def build_request_from_body(self, message_body: dict) -> BoltRequest:
+        timestamp, body = str(int(time.time())), json.dumps(message_body)
+        return BoltRequest(body=body, headers=self.build_headers(timestamp, body))
+
+    def test_success(self):
```
seratch, line 55: Can you add an extra test that verifies the decorator use cases? In my understanding, the following code should work too. If it does not work for some reason, I can explore possible ways too.
```python
func = app.function("c")
@func 
def main_func():
    pass

@func.action("a")
def handle_actions():
    pass
```

seratch (changes requested):
@WilliamBergamin Great work. Can you check these points?

slack_bolt/util/utils.py
WilliamBergamin: Agreed, changed to `extract_listener_callables`

slack_bolt/context/async_context.py
WilliamBergamin: Good point, I ended up moving this information to `slack_bolt/context/complete/complete & async_complete` in order for the doc string to be visible in IDE previews. (tested with vs code)

tests/scenario_tests/test_function_block_action.py
WilliamBergamin: @seratch I've added the test but it does not work exactly like the mentioned use case, here is the use case that currently works
```python
func = app.function("c")
@func 
def main_func():
    pass

@main_func.action("a")
def handle_actions():
    pass
```

This is the behavior I expected since `app.function` returns a `function` that then returns a `SlackFunction object` when called

From what I understand in order to implement the above use case the `SlackFunction object` would need to be returned by `app.function` and return itself when `SlackFunction object.__call__()` is used, this might go against the behavior of the other app decorators that return the `registered listener`.

Currently `SlackFunction object.__call__()` executes the `registered listener` and returns the output, `SlackFunction` tries to wrap the registered listener with an object that has additional decorators and  stores the callback_id.

slack_bolt/slack_function/slack_function.py
WilliamBergamin: @seratch By adding the middleware globally won't this make it execute every time a new request comes in? My intended behavior was to have this middleware execute only when the request matches the `function` type listeners

WilliamBergamin: Maybe creating a middleware for a subgroup of listeners could be something worth exploring

seratch: @WilliamBergamin 
> By adding the middleware globally won't this make it execute every time a new request comes in?

Yes, it can be but it's totally fine. If a request does not have the slack function token, the middleware does not do anything. The overhead is ignorable. More importantly, the middleware can be executed **only once** in any case. Contrarily, if we go with listener middleware for this and a developer has 10 functions in an app, the same middleware can be executed 10 times at maximum. This is because listener middleware can be executed every time bolt matches a request with a listener's matchers/middleware.

>My intended behavior was to have this middleware execute only when the request matches the function type listeners

In addition to above, if you have function listener at top of the list of middleware, the listener matcher can be executed for all incoming requests. Again, Bolt tries to find a listener to run for an incoming request. During this process, all the listener matchers/middleware can be executed for all requests.

So, in this case, we should go with a global middleware in the interest of simplicity and consistently small overhead.

seratch (changes requested):

docs/api-docs/slack_bolt/cli/get_hooks.html
```diff
@@ -0,0 +1,97 @@
+<!doctype html>
```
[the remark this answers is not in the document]
WilliamBergamin: Done, this was added by accident

slack_bolt/slack_function/slack_function.py
WilliamBergamin: This has been implemented, thanks for explaining I initially thought this middleware was executed after the match occurred 💯

tests/scenario_tests/test_function_block_action.py
seratch: @WilliamBergamin Using the `main_func` method as a decorator object should be quite surprising for developers. I understand that the current implementation can work in this way, but I don't think we should go with this.

tests/scenario_tests/test_function_decorators.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def test_decorators(self):
+        app = App(signing_secret=self.signing_secret, client=self.web_client)
+        ack = NoopAck()
+
+        @app.function("c")
+        def handle_function_events(body: dict):
+            assert body is not None
+
+        @handle_function_events.action("some-func-action-id")
```
seratch, line 40: We should not go with this. Using the above method reference as a decorator object like this is not a common practice at all.

WilliamBergamin: This pattern came from the [functools library](https://docs.python.org/3/library/functools.html#functools.singledispatchmethod) where it seems to appear in few places, The pattern seems to fit our use case well it is true that it does not appear in other web frameworks but I don't think the use case is very common

seratch: @WilliamBergamin Ah, I didn't know these design in the standard libraries. Although I still feel this design can be confusing for some developers including me (investigating the reason why it does not work can be hard), I can agree to go with this if you have confidence on this.

WilliamBergamin: I see how details of the pattern can be missed, since I've used this pattern in the past I may be more familiar with it, I will ask other members of the team for their opinion on this 🧠 

but overall, I do feel confident with this implementation

WilliamBergamin: In this [PR](https://github.com/WilliamBergamin/bolt-python/pull/2) I've implemented a prototype that allows the use case described[ here](https://github.com/slackapi/bolt-python/pull/715#discussion_r961157041) to work without breaking the existing behavior

seratch: Thanks, it looks great to me

seratch (approved):

PR #718, opened 2022-09-13
[future beta] Adding remote function view interactivity

WilliamBergamin:
This PR aims to provide remote function `view` interactivity capabilities and follows the same approach as `action` interactivities.

Refer to [theses steps in order to test locally](https://github.com/slackapi/bolt-python/pull/715#issue-1357832231)

### Category 

* [ ] Others

## Requirements

### DIFF
--- a/slack_bolt/listener_matcher/builtins.py
+++ b/slack_bolt/listener_matcher/builtins.py
@@ -293,8 +293,9 @@ def function_view_closed(
 ) -> Union[ListenerMatcher, "AsyncListenerMatcher"]:
     def func(body: Dict[str, Any]) -> bool:
         return (
-            _function_interactivity_match(function_callback_id, body),
-            is_view_closed(body) and _matches(callback_id, body["view"]["callback_id"]),
+            _function_interactivity_match(function_callback_id, body)
+            and is_view_closed(body)
+            and _matches(callback_id, body["view"]["callback_id"])
         )
 
     return build_listener_matcher(func, asyncio, base_logger)
--- a/slack_bolt/listener_matcher/builtins.py
+++ b/slack_bolt/listener_matcher/builtins.py
@@ -259,12 +259,12 @@ def function_view(
     base_logger: Optional[Logger] = None,
 ) -> Union[ListenerMatcher, "AsyncListenerMatcher"]:
     if isinstance(constraints, (str, Pattern)):
-        return function_view_submission(callback_id, constraints, asyncio)
+        return function_view_submission(callback_id, constraints, asyncio, base_logger)
     elif "type" in constraints:
         if constraints["type"] == "view_submission":
-            return function_view_submission(callback_id, constraints["callback_id"], asyncio)
+            return function_view_submission(callback_id, constraints["callback_id"], asyncio, base_logger)
         if constraints["type"] == "view_closed":
-            return function_view_closed(callback_id, constraints["callback_id"], asyncio)
+            return function_view_closed(callback_id, constraints["callback_id"], asyncio, base_logger)
 
     raise BoltError(f"function view ({constraints}: {type(constraints)}) must be any of str, Pattern, and dict")
 
--- a/slack_bolt/listener_matcher/builtins.py
+++ b/slack_bolt/listener_matcher/builtins.py
@@ -209,19 +209,20 @@ def function_event(
 # functions
 
 
+def _function_interactivity_match(callback_id: str, body: Dict[str, Any]) -> bool:
+    return is_function_interactivity(body) and _matches(
+        callback_id, body.get("function_data", {}).get("function", {}).get("callback_id", "")
+    )
+
+
 def function_action(
-    callback_id: Union[str, Pattern],
+    callback_id: str,
     constraints: Union[str, Pattern, Dict[str, Union[str, Pattern]]],
     asyncio: bool = False,
     base_logger: Optional[Logger] = None,
 ) -> Union[ListenerMatcher, "AsyncListenerMatcher"]:
     if isinstance(constraints, (str, Pattern)):
-
-        def func(body: Dict[str, Any]) -> bool:
-            return _function_block_action(callback_id, constraints, body)
-
-        return build_listener_matcher(func, asyncio, base_logger)
-
+        return function_block_action(callback_id, constraints, asyncio)
     elif "type" in constraints:
         action_type = constraints["type"]
         if action_type == "block_actions":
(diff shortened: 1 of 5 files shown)


slack_bolt/listener_matcher/builtins.py
```diff
@@ -235,28 +236,66 @@ def func(body: Dict[str, Any]) -> bool:
...
+            and _does_block_id_match(constraints, body)
+        )
 
-    if _matches(callback_id, body.get("function_data", {}).get("function", {}).get("callback_id", "")) is False:
-        return False
+    return build_listener_matcher(func, asyncio, base_logger)
 
-    return _does_block_id_match(constraints, body)
 
+def function_view(
```
WilliamBergamin, line 255: Thinking of refactoring this and the `view` function since the code is similar in order reduce code duplication 💡 
Maybe a factory method pattern could work, but it will definitely increase complexity

slack_bolt/slack_function/slack_function.py
```diff
@@ -78,8 +77,95 @@ def __call__(*args, **kwargs):
...
+
+    def view(
+        self,
+        constraints: Union[str, Pattern, Dict[str, Union[str, Pattern]]],
+        matchers: Optional[Sequence[Callable[..., bool]]] = None,
+        middleware: Optional[Sequence[Union[Callable, Middleware]]] = None,
+    ) -> Callable[..., Optional[Callable[..., Optional[BoltResponse]]]]:
+        """Registers a new `view_submission`/`view_closed` event listener for a function.
+
+            # Use this method as a decorator
```
filmaj, line 92: "This method" is probably a copy-paste from the other docstring for the top-level function handler, yes? Probably want to rename this to something else. Maybe:
```suggestion
            # Define a function handler using the function decorator
```

slack_bolt/slack_function/slack_function.py
```diff
@@ -78,8 +77,95 @@ def __call__(*args, **kwargs):
...
+            @app.function("function_1")
+            def sample_view(event, complete: Complete):
+                client.views_open(
+                    interactivity_pointer=interactivity_pointer,
```
filmaj, line 96: Do you think we need to explain in here where a user can get an interactivity pointer from?

slack_bolt/slack_function/slack_function.py
```diff
@@ -78,8 +77,95 @@ def __call__(*args, **kwargs):
...
+                    view={...}
+                )
+
+            # Use this method as a decorator
+            @my_function.view("view_1")
+            def handle_submission(ack, body, client, view, complete):
+                # Acknowledge the view_submission event and close the modal
```
filmaj, line 103: In this specific example, how is the `handle_submission` function only applying to view_submission events?

slack_bolt/slack_function/slack_function.py
```diff
@@ -78,8 +77,95 @@ def __call__(*args, **kwargs):
...
+                # Validate the inputs
+                errors = {}
+                if hopes_and_dreams is not None and len(hopes_and_dreams) <= 5:
+                    errors["block_c"] = "The value must be longer than 5 characters"
+                if len(errors) > 0:
+                    ack(response_action="errors", errors=errors)
+                    return
+                # complete the function
+                complete()
+                # Do whatever you want with the input data - here we're saving it to a DB
```
filmaj, line 117: This line probably does not apply

slack_bolt/slack_function/slack_function.py
```diff
@@ -78,8 +77,95 @@ def __call__(*args, **kwargs):
...
+
+            # Pass a function to this method
+            sample_view_func = app.function("request-approval")(sample_view)
+            sample_view_func.action("view_1")(handle_submission)
```
filmaj, line 121: Should this be `action` or `view`?

slack_bolt/slack_function/slack_function.py
```diff
@@ -78,8 +77,95 @@ def __call__(*args, **kwargs):
...
+
+        return __call__
+
+    def view_submission(
+        self,
+        constraints: Union[str, Pattern],
+        matchers: Optional[Sequence[Callable[..., bool]]] = None,
+        middleware: Optional[Sequence[Union[Callable, Middleware]]] = None,
+    ) -> Callable[..., Optional[Callable[..., Optional[BoltResponse]]]]:
+        """Registers a new `view_submission` listener."""
```
filmaj, line 145: Would devs use the `view` method above or these more specific `view_submission` / `view_closed` methods?

filmaj (commented):
Left a few questions and comments, nicely done!

slack_bolt/slack_function/slack_function.py
WilliamBergamin: Added this to the function.actions handler as well 🥇

slack_bolt/slack_function/slack_function.py
WilliamBergamin: I feel like this should be written in the docs and we should add a reference to it here (but I don't think the docs are written for this yet) I've had some average experiences dealing with very lengthy suggestion boxes in IDEs

slack_bolt/slack_function/slack_function.py
WilliamBergamin: This follows the same behavior as a `global view interactivity listener` if a String is passed in here then Bolt assumes this is the `callback_id` for a `view_submission` event

slack_bolt/slack_function/slack_function.py
WilliamBergamin: Definitely `view` good catch

slack_bolt/slack_function/slack_function.py
WilliamBergamin: Based on the [existing doc](https://slack.dev/bolt-python/concepts#view_submissions) for view_submission we recommend using `.view("my string")` to set up a `view_submission` listener, we don't have a dedicated section for `view_closed`

but here are some tested ways to set up `view_closed`
```python
app.view({"type": "view_closed", "callback_id": "view-id"})(simple_listener)
func.view_closed("view-id")(simple_listener)
```

filmaj: Ohhh that's right! Makes sense, and thanks for the explainer.

slack_bolt/slack_function/slack_function.py
```diff
@@ -46,21 +46,20 @@ def action(
     ) -> Callable[..., Optional[Callable[..., Optional[BoltResponse]]]]:
         """Registers a new action listener to your function. This method can be used as either a decorator or a method.
 
-            # Use this method as a decorator
+            # Define a function handler using the function decorator
             @app.function("request-approval")
             def request_approval(event, complete: Complete):
-                complete(outputs={})
+                // do something
```
seratch, line 52: // line comment does not work in Python
```suggestion
                # do something
```

slack_bolt/slack_function/slack_function.py
seratch: hm, I think the code snippet here should show a bit more realistic code. So, demonstrating how to extract interactivity_pointer from event data would be preferred.

slack_bolt/slack_function/slack_function.py
```diff
@@ -78,8 +77,94 @@ def __call__(*args, **kwargs):
...
+                client.views_open(
+                    interactivity_pointer=interactivity_pointer,
+                    view={...}
+                )
+
+            # Use this method as a decorator
+            @my_function.view("view_1")
+            def handle_submission(ack, body, client, view, complete):
+                # Acknowledge the view_submission event and close the modal
+                ack()
```
seratch, line 104: You need to move this line to the part below the input validation part. Meaning that the current code snippet calls `ack()` twice in the case of input errors.

seratch (changes requested):
@WilliamBergamin Great work. Are you going to work on the asyncio side in a different PR?

slack_bolt/slack_function/slack_function.py
WilliamBergamin: Seems like my java reflexes kicked in here 😄

WilliamBergamin:
@seratch Exactly asyncio side will be in a different PR

slack_bolt/slack_function/slack_function.py
WilliamBergamin: I've added how to extract interactivity_pointer 💯

slack_bolt/slack_function/slack_function.py
```diff
@@ -78,8 +77,97 @@ def __call__(*args, **kwargs):
...
+            def sample_view(event, complete: Complete):
+                # Assume there's an interactivity object passed in as input `interactivity`
+                interactivity_pointer = event["inputs"]["interactivity"]["interactivity_pointer"]
+                client.views_open(
+                    interactivity_pointer=interactivity_pointer,
+                    view={...}
+                )
+
+            # Use this method as a decorator
+            @my_function.view("view_1")
```
seratch, line 103:
```suggestion
            @ sample_view.view("view_1")
```

WilliamBergamin: Good catch, this was a typo 💯

seratch (approved):

PR #722, opened 2022-09-20
Fix #721 Passing a global dict object without channel prop can cause issues among requests

### LINKED ISSUE #721: Passing a global dict object without channel prop can cause issues among requests
I did try to answer as much as I though was relevant below, but I found the exact line which is leading to the bug. 

https://github.com/slackapi/bolt-python/blob/main/slack_bolt/context/say/say.py#L51

I am passing a global dict to SAY function and that is getting modified here. 
Next time the SAY function is called with the same dictionary, it ignores the current channel and send the message to the very first user. 

Please consider using ".copy()" or "copy.copy" method on this line,
https://github.com/slackapi/bolt-python/blob/main/slack_bolt/context/say/say.py#L49

https://stackoverflow.com/questions/2465921/how-to-copy-a-dictionary-and-only-edit-the-copy

### Reproducible in:
Pass a global dict as response to say(). 

#### The `slack_bolt` version
latest

#### Python runtime version
3.9

#### OS info
windows 10

### Expected result:
Say should reply the global reply dict to the user who started conversation. 
Not, to the very fist user who used the service endpoint for the first time.

seratch:
This pull request resolves #721 

* [ ] Others

### DIFF
--- a/slack_bolt/context/say/async_say.py
+++ b/slack_bolt/context/say/async_say.py
@@ -1,6 +1,7 @@
 from typing import Optional, Union, Dict, Sequence
 
 from slack_bolt.context.say.internals import _can_say
+from slack_bolt.util.utils import create_copy
 from slack_sdk.models.attachments import Attachment
 from slack_sdk.models.blocks import Block
 from slack_sdk.web.async_client import AsyncWebClient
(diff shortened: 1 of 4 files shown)


WilliamBergamin (approved):
This looks good to me :shipit: could we have used a shallow copy?

seratch:
@WilliamBergamin Thanks for your review! Indeed, using a shallow copy works as of today but for safety with any future changes, I prefer a deep copy this time.

PR #736, opened 2022-10-09
Add context 'user_id' extraction for 'message_changed' and 'message_deleted' events 

eddyg:
I noticed that `context.user_id` was not being set for `message_changed` or `message_deleted` `message` subtype events. `user` is located in _payload_ → _event_ → _message_ for `message_changed` events, and in _payload_ → _event_ → _previous_message_ for `message_deleted` events.

The first and third commits are for `message_changed` (with tests); the fourth commit is for `message_deleted` (with tests).

The second commit addresses a typing issue I noticed by ensuring that `in` is not being compared to `None` (the original code would generate _"in" not supported for type "None"_ warnings. I also removed the `# type: ignore` comments on the associated `return` statements as they don't seem to be required.

* [X] `slack_bolt.async_app.AsyncApp` and/or its core components
* [ ] Others

### DIFF
--- a/slack_bolt/request/internals.py
+++ b/slack_bolt/request/internals.py
@@ -113,6 +113,9 @@ def extract_user_id(payload: Dict[str, Any]) -> Optional[str]:
         return payload.get("user_id")
     if payload.get("event") is not None:
         return extract_user_id(payload["event"])
+    if payload.get("message") is not None:
+        # message_changed: body["event"]["message"]
+        return extract_user_id(payload["message"])
     return None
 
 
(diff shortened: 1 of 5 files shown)


seratch:
Hi @eddyg, thanks for sending this PR! Could you add the unit tests covering the pattern? Adding new files like the following would be easier to maintain for us. Please note that we need tests for both `App` and `AsyncApp`. Thanks!

* https://github.com/slackapi/bolt-python/blob/v1.15.1/tests/scenario_tests/test_message_thread_broadcast.py
* https://github.com/slackapi/bolt-python/blob/v1.15.1/tests/scenario_tests_async/test_message_thread_broadcast.py

eddyg:
> Hi @eddyg, thanks for sending this PR! Could you add the unit tests covering the pattern? Adding new files like the following would be easier to maintain for us. Please note that we need tests for both `App` and `AsyncApp`. Thanks!

Alright, I took a shot at a simple test case (including an async version) to verify `context.user_id` in `message_changed` events...

seratch:
@eddyg Thanks for adding tests! I have a few things to verify before merging this. I will work on it early next week.

PR #758, opened 2022-11-04
Fix #754 by adding the async version of Tornado adapter

### LINKED ISSUE #754: Does the Tornado adapter support AsyncApp?
Hello, I am trying to set up a slack app using the Tornado adapter. I am trying to use AsyncApp instead of App, as that is the purpose of Tornado as an Asyncio server. however, it looks like the Tornado adapter tries to call `self.app.dispatch` and doesn't attempt to call `await self.app.async_dispatch` which is what the AsyncApp class uses. CC @seratch (Saw that you had worked on the Tornado code)

I am successfully able to get it working with **App** as seen below:

```
slack_app = App(
    token=config.get("_global_.secrets.slack.bot_token"),
    signing_secret=config.get("_global_.secrets.slack.signing_secret")
)
.... (routes):
("/api/v3/slack/events", SlackEventsHandler, dict(app=slack_app)),
("/api/v3/slack/install", SlackOAuthHandler, dict(app=slack_app)),
("/api/v3/slack/oauth_redirect", SlackOAuthHandler, dict(app=slack_app)),
```

but I am unable to get this working with an **AsyncApp**. The app starts correctly but when I receive events, I get an error like this:

```
HTTPServerRequest(protocol='http', host='REDACTED', method='POST', uri='/api/v3/slack/events', version='HTTP/1.1', remote_ip='127.0.0.1')
Traceback (most recent call last):
  File "env/lib/python3.10/site-packages/tornado/web.py", line 1711, in _execute
    result = method(*self.path_args, **self.path_kwargs)
  File "env/lib/python3.10/site-packages/slack_bolt/adapter/tornado/handler.py", line 17, in post
    bolt_resp: BoltResponse = self.app.dispatch(to_bolt_request(self.request))
AttributeError: 'AsyncApp' object has no attribute 'dispatch'
2022-11-01:16:26:54,418 ERROR    [web.py:2271] 500 POST /api/v3/slack/events (127.0.0.1) 34.36ms
2022-11-01:16:26:55,176 ERROR    [web.py:1798] Uncaught exception POST /api/v3/slack/events (127.0.0.1)
```

### Reproducible in:

slack-bolt==1.15.2
slack-sdk==3.19.2
Python 3.10.8

#### OS info

#58-Ubuntu SMP Thu Oct 13 08:03:55 UTC 2022

(Ubuntu 22.04)

seratch:
This pull request resolves #754 by adding the async version of Tornado adapter. Refer to #754 for the context and details.

* [ ] Others

### DIFF
--- a/examples/tornado/async_app.py
+++ b/examples/tornado/async_app.py
@@ -2,28 +2,28 @@ import logging
 
 logging.basicConfig(level=logging.DEBUG)
 
-from slack_bolt import App
-from slack_bolt.adapter.tornado import SlackEventsHandler
+from slack_bolt.async_app import AsyncApp
+from slack_bolt.adapter.tornado.async_handler import AsyncSlackEventsHandler
 
-app = App()
+app = AsyncApp()
 
 
 @app.middleware  # or app.use(log_request)
-def log_request(logger, body, next_):
+async def log_request(logger, body, next_):
     logger.debug(body)
-    next_()
+    await next_()
 
 
 @app.event("app_mention")
-def event_test(body, say, logger):
+async def event_test(body, say, logger):
     logger.info(body)
-    say("What's up?")
+    await say("What's up?")
 
 
 from tornado.web import Application
 from tornado.ioloop import IOLoop
 
-api = Application([("/slack/events", SlackEventsHandler, dict(app=app))])
+api = Application([("/slack/events", AsyncSlackEventsHandler, dict(app=app))])
 
 if __name__ == "__main__":
     api.listen(3000)
@@ -32,4 +32,4 @@ if __name__ == "__main__":
 # pip install -r requirements.txt
 # export SLACK_SIGNING_SECRET=***
 # export SLACK_BOT_TOKEN=xoxb-***
-# python app.py
+# python async_app.py
--- a/examples/tornado/async_oauth_app.py
+++ b/examples/tornado/async_oauth_app.py
@@ -1,21 +1,21 @@
 import logging
-from slack_bolt import App
-from slack_bolt.adapter.tornado import SlackEventsHandler, SlackOAuthHandler
+from slack_bolt.async_app import AsyncApp
+from slack_bolt.adapter.tornado.async_handler import AsyncSlackEventsHandler, AsyncSlackOAuthHandler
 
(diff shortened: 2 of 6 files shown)


examples/tornado/app.py
```diff
@@ -9,9 +9,9 @@
 
 
 @app.middleware  # or app.use(log_request)
-def log_request(logger, body, next):
+def log_request(logger, body, next_):
```
seratch, line 12: just a minor improvement to eliminate warnings by IDE. See also: https://github.com/slackapi/bolt-python/pull/394

WilliamBergamin (changes requested):
Great work on this, the code looks good to me 🥇

I ran some test locally that modify the behavior in order to deploy a `sync App` on the `asyncTornado` adapter and this seemed to work 😮 I believe the async handler just executes the sync code as a block similarly to how it would `await` for the `async_dispatch` 

 I'm not sure if this is behavior we want in this adapter, but I think it could be possible for this adapter to accept `async App` and `sync Apps`

seratch:
@WilliamBergamin Sorry, I don't understand your point yet. Can you elaborate more on the changes you are requesting in your comment?

The intention to add a new adapter in this PR is to enable developers to use AsyncApp along with Tornado. Since AsyncApp relies on AIOHTTP, we need to have a separate module for it. Also, AsyncApp runs "async" functions for handling requests from Slack. This is completely different from the App approach. Let me know if I am missing something here.

seratch:
>I'm not sure if this is behavior we want in this adapter, but I think it could be possible for this adapter to accept async App and sync Apps

I clearly didn't answer this but indeed it's technically possible to accept both App and AsyncApp in this new adapter. However, perhaps no one will use it. Also, even when a single handler supports both, two separate handlers should exist for sync App users who do not need AIOHTTP at all. Again, just importing AsyncApp requires AIOHTTP installation, which is optional dependency of slack-bolt/slack-sdk. In addition, for consistency with other built-in adapters, I prefer supporting only AsyncApp in this case.

Let's discuss more on this next week.

WilliamBergamin:
@seratch this makes a lot of sense thank you for the answer 💯

WilliamBergamin (approved):
This looks good to me 💯

PR #776, opened 2022-11-28
CI python 3.6 bug fix

WilliamBergamin:
From [this issue](https://github.com/actions/setup-python/issues/544) it was found that the ubuntu `latest-version` of the github action runner was updated to 22.04, there are no tarballs in 22.04 for any python 3.6.* versions.

This PR introduces a fix to allow us to run the python 3.6 test without issues.

### Category

* [x] Others

## Requirements

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -7,16 +7,12 @@ on:
 
 jobs:
   build:
-    runs-on: ${{ matrix.os }}
+    # Avoiding -latest due to https://github.com/actions/setup-python/issues/162
+    runs-on: ubuntu-20.04
     timeout-minutes: 10
     strategy:
       matrix:
-        os:
-          - ubuntu-latest
-        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
-        include:
-          - python-version: "3.6"
-            os: ubuntu-20.04
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10", "3.11"]
     env:
       # default: multiprocessing
       # threading is more stable on GitHub Actions
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -11,6 +11,8 @@ jobs:
     timeout-minutes: 10
     strategy:
       matrix:
+        os:
+          - ubuntu-latest
         python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
         include:
           - python-version: "3.6"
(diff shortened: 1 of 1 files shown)


WilliamBergamin:
For some reason the Github runner is using `ubuntu 22.04` instead of the correct os version of `ubuntu 20.04` to run tests with python 3.6 refer to [the logs](https://github.com/slackapi/bolt-python/actions/runs/3568683511/jobs/5997984647#step:1:8)

[On my fork](https://github.com/WilliamBergamin/bolt-python/actions/runs/3568610012) this seems to work as expected not  sure why the runner is using the latest version here

seratch:
@WilliamBergamin Usually testing with the latest is recommended, as we don't need to do tests with any specific OS distribution and version. I also observed this issue caused by the behavior change in the GitHub Actions infra side for python-slack-sdk project. Here is my fix for it: https://github.com/slackapi/python-slack-sdk/commit/f2f3fd63e7b86abdb29a66f28b897402802ddafe I am not sure why your change does not work here, but if my change works well for bolt-python, applying the same change may help.

.github/workflows/tests.yml
```diff
@@ -7,11 +7,16 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
```
WilliamBergamin, line 10: @seratch I was not specifying the matrix os version as the `run-on` value, let me know if you want to use this approach which default to `ubuntu-latest` or the one found in the [python-sdk](https://github.com/slackapi/python-slack-sdk/commit/f2f3fd63e7b86abdb29a66f28b897402802ddafe)

seratch:
@WilliamBergamin It seems that, with your change, we need to adjust the main branch protection setting every time we change the OS image setting. Meaning that "build (3.10)" and "build (ubuntu-latest, 3.10)" are non-equivalent. You need to head to https://github.com/slackapi/bolt-python/settings and then go to "Code and automation" > "Branches" > "Branch protection rules" > Edit "main" branch > Adjust the list for "Require branches to be up to date before merging". 

Considering that, simply changing the runs-on setting for all looks like easier to maintain to me. As I said above, we don't have any requirements for the OS version. Using -latest is just a recommendation by Github for stability in long-term maintenance. What do you think?

WilliamBergamin:
@seratch I agree think we can fix the os version and update it manually, this is the only place I've seen recommend using the `latest` tag, I've been told it leads to unexpected dependency errors like the one we are encountering

seratch (approved):

PR #939, opened 2023-07-30
[Django 4.x] Fix never returning an existing installation or bot

chpmrc:
The `installed_at` field should not be part of the filter, or we would never return an existing, older installation.

This fixes a recurring `MultipleObjectsReturned` error that whoever uses this example will face.

* [x] Others (Django examples)

### DIFF
--- a/examples/django/oauth_app/slack_datastores.py
+++ b/examples/django/oauth_app/slack_datastores.py
@@ -42,7 +42,6 @@ class DjangoInstallationStore(InstallationStore):
             SlackInstallation.objects.filter(client_id=self.client_id)
             .filter(enterprise_id=installation.enterprise_id)
             .filter(team_id=installation.team_id)
-            .filter(installed_at=i["installed_at"])
             .first()
         )
         if row_to_update is not None:
--- a/examples/django/oauth_app/slack_datastores.py
+++ b/examples/django/oauth_app/slack_datastores.py
@@ -66,7 +66,6 @@ class DjangoInstallationStore(InstallationStore):
             SlackBot.objects.filter(client_id=self.client_id)
             .filter(enterprise_id=bot.enterprise_id)
             .filter(team_id=bot.team_id)
-            .filter(installed_at=b["installed_at"])
             .first()
         )
         if row_to_update is not None:


examples/django/oauth_app/slack_datastores.py
```diff
@@ -42,7 +42,6 @@ def save(self, installation: Installation):
             SlackInstallation.objects.filter(client_id=self.client_id)
             .filter(enterprise_id=installation.enterprise_id)
             .filter(team_id=installation.team_id)
-            .filter(installed_at=i["installed_at"])
```
seratch, line 45: When we remove this part, perhaps, adding `order by installed_at desc` is necessary to make the query result consistent. What do you think?

examples/django/oauth_app/slack_datastores.py
```diff
@@ -66,7 +65,6 @@ def save_bot(self, bot: Bot):
             SlackBot.objects.filter(client_id=self.client_id)
             .filter(enterprise_id=bot.enterprise_id)
             .filter(team_id=bot.team_id)
-            .filter(installed_at=b["installed_at"])
```
seratch, line 69: same as above

examples/django/oauth_app/slack_datastores.py
chpmrc: Hey @seratch there might be some ambiguity here:
- Are we keeping track of every installation? As in: if I uninstall/reinstall the app I want 2 entries in the DB, not 1?
- Or are we only storing a single entry per workspace?

In the former case then I'm not sure how "updating" existing rows would work. Is there a case where the oauth redirect values could be changing for an installation at a specific timestamp (`installed_at`)?

In the latter then it wouldn't make sense to store a timestamp since it will change any time the app is reinstalled into the workspace.

So to answer your question: `order by installed_at desc` would make sense if we were looking up the most recent installation (case 1) but wouldn't make any difference for case 2.

seratch: >Are we keeping track of every installation? 

Yes. If your code calls the save method outside of the installation process, the save method may update an existing data. So, both inserting and updating a row are possible scenario.

>As in: if I uninstall/reinstall the app I want 2 entries in the DB, not 1?
Or are we only storing a single entry per workspace?

The example code there does not support uninstallation scenario since the method was not available back then. This can be improved too (perhaps in a different PR is easier to manage).

While checking your change again, I came to think that, if you encounter a runtime exception in some scenario, resolving it could be a different approach such as catching the exception and switching to insert operation anyways. What do you think?

chpmrc: This isn't about catching an exception, it's about understanding what this piece of code does. My understanding is that this is simulating `update_or_create` https://docs.djangoproject.com/en/4.2/ref/models/querysets/#update-or-create, which would make adding `installed_at` in the filter logically wrong.

If that's not the case I'd like to understand when it might happen that we receive updated fields for an existing installation at a specific timestamp.

By removing these lines we ensure that the same app can be (re)installed (not necessarily after uninstalling it but, e.g., because the scope or some other attributes changed) as many times as the user wants while only keeping a single entry that represents the most recent installation.

You can try this pretty easily: clone the example, fill out the required env variables, install the app (and get the oauth callback), then try reinstalling the app and you will get an `IntegrityError` because you'll be inserting a new installation, with the same `client_id, enterprise_id, team_id, user_id` but a different `installed_at`, which violates the index on the `Installation` model.

seratch: Thanks for sharing that. The error never happened with Django 3.x and as you can see in the `requirements.txt`, it supports 3.x only right now (right, it's about time to upgrade it). So, perhaps the code needs to be updated for Django 4.x compatibility. Since it's already Saturday for me (I'm in APAC timezone), let me check more details early next week.

seratch: @chpmrc I've verified that the current code works without any issues (tested with Django 4.2.4). The query to find existing data can return a row only when your app enables token rotation. Otherwise, Bolt always inserts a new row for every single installation. I am still unsure about the steps to reproduce your situation, but I hope you'll figure out a solution for yourself. By the way, I am on PTOs this week, so I'll be slow to respond for a while.

seratch:
As I mentioned at https://github.com/slackapi/bolt-python/pull/939#discussion_r1285323417, I didn't manage to reproduce the situation you shared. Let us close this PR for better task management but if you have comments please feel free to write in. Thanks again for submitting this PR!

examples/django/oauth_app/slack_datastores.py
chpmrc: Ok I've tested the example in an isolated environment and it is indeed creating multiple installations, which is what this was all about. I was expecting this `save` to find an existing installation (for the same enterprise, team, etc.) and just update it. What's the rationale behind creating multiple installations though? Thanks!

seratch: @chpmrc When you're managing only bot installations, maintaining a single database row for each one is usually sufficient. However, the Slack platform also supports user-level OAuth permission grants. In such scenarios, each installation can be associated with a different user. For this reason, we recommend saving each installation as a new record. 

If you prefer to minimize historical data even with user installations, you can manage the data on a per 'team_id + user_id' basis by upserting rows. If you manage only bot installations, the code can be modified to be simpler. I hope this clarifies.

chpmrc: Got it, I clarified above why I would not expect multiple installations to be created but your point about user level permissions makes sense. Thanks for taking the time to clarify all that!

PR #986, opened 2023-11-16
Add remote function support

WilliamBergamin:
This PR aims to add minimalist support for remote functions in python

### Feedback

I'm looking for feedback on
- Usability, is this what we envisioned the developer experience would be?
- Does this aline with the Bolt-js experience?
- Implementation, does this implementation align with bolt principles
- Tests, is this missing tests that could be valuable for the feature

### Testing

I created [this gist](https://gist.github.com/WilliamBergamin/e7ccc6ad4383b6883f832cbef2b6215a) to show how to test this new feature, use it for the following steps

1. Create a new app
    a. Head to https://api.slack.com/apps/?new_app=1
    b. Click "From an app manifest"
    c. Select YAML and paste the above manifest.yml content
    d. Click create button
2. Install the app into org/workspace
    a. Install the app anyways (when the workspace is in an Org, org-wide installation is required to publish functions)
    b. Grab the xoxb- token (Settings > Install App > Bot User OAuth Token)
    c. `export SLACK_BOT_TOKEN=xoxb-...`
3. Set up Socket Mode
    a. Head to Settings > Basic Information > App-Level Tokens on the https://api.slack.com/apps page
    b. Generate a new token with connections:write scope
    c. `export SLACK_APP_TOKEN=xapp-...`
4. Spin up the app
    a. Add `app.py` and `async_app.py` with the above source code
    b. Pull the latest PR branch of bolt python to your local
    c. Run `scripts/build_pypi_package.sh` this will generate a `.whl` in the `dist/` folder
    d. (Optional) In your project set up a venv with `python -m venv .venv` and `source .venv/bin/activate`
    e. `pip install global/path/to/bolt-python/dist/something.whl`
    f. `pip install aiohttp`
    g `python app.py` or `python async_app.py`
5. Add the custom step to a workflow
    a. Open the workflow builder
    b. Create a new workflow with a link trigger
    d. Seach "Hello" in the Steps view on the right side
    e. Add the step to the workflow (you can set any user ID as the input)
    f. Publish the workflow

### Category

* [ ] Others

## Requirements


slack_bolt/logger/messages.py
```diff
@@ -266,7 +267,26 @@ def warning_unhandled_request(  # type: ignore
...
+            # @app.function
+            callback_id = event.get("function", {}).get("callback_id", "function_id")
+            return _build_unhandled_request_suggestion(
+                default_message,
+                f"""
+@app.function("{callback_id}")
+{'async ' if is_async else ''}def handle_{callback_id}_function(body, event, client, logger):
+    logger.info(body)
+    {'await ' if is_async else ''}client.api_call(
+        "functions.completeError",
```
seratch, line 282: why having the error  handler only? Generating something like this would be even more developer-friendly:
```python
@app.function("{callback_id}")
def handle_{callback_id}_function_events(body, event, client, logger):
    try:
        # TODO: do something here
        outputs = {}
        client.api_call(
            "functions.completeSuccess",
            json={
                "function_execution_id": context.function_execution_id,
                "outputs": outputs,
            },
        )
    except Exception as e:
        error = f"Failed to handle a function request (error: {e})"
        client.api_call(
            "functions.completeError",
            json={
                "function_execution_id": context.function_execution_id,
                "error": error,
            },
        )
```

slack_bolt/request/async_internals.py
```diff
@@ -43,6 +45,12 @@ def build_async_context(
         context["channel_id"] = channel_id
     if "response_url" in body:
         context["response_url"] = body["response_url"]
+    function_execution_id = extract_function_execution_id(body)
```
seratch, line 48: Adding these lines here can break the request_url vs request_urls behavior (their if + elif statement coalation). can you move these  to somewhere else?

slack_bolt/request/internals.py
```diff
@@ -208,6 +208,26 @@ def extract_channel_id(payload: Dict[str, Any]) -> Optional[str]:
...
 
+def extract_function_execution_id(payload: Dict[str, Any]) -> Optional[str]:
+    if payload.get("function_execution_id") is not None:
+        return payload.get("function_execution_id")
+    if payload.get("event") is not None:
+        return extract_function_execution_id(payload["event"])
+    if payload.get("function_data") is not None:
+        return extract_function_execution_id(payload["function_data"])
+    if payload.get("execution_id") is not None:
+        return payload.get("execution_id")
```
seratch, line 219: The more general term "execution_id" can be used in a different menaing in the future (say, as part of an Events API payload). So avoiding the recursive calls for it would be safer:
```suggestion
    if payload.get("function_data") is not None:
        return payload["function_data"].get("execution_id")
```

slack_bolt/request/internals.py
```diff
@@ -208,6 +208,26 @@ def extract_channel_id(payload: Dict[str, Any]) -> Optional[str]:
...
+    return None
+
+
+def extract_function_bot_access_token(payload: Dict[str, Any]) -> Optional[str]:
+    if payload.get("bot_access_token") is not None:
```
seratch, line 224: Since this key is not used in any other paterns, this simple implementation is fine so far. If we want to make it safer, checking the existence of either function_execution_id or function_data in the same dict object would be worth considering. That said, this suggestion is totally optional.

tests/scenario_tests/test_function.py
```diff
@@ -0,0 +1,242 @@
+import json
```
seratch, line 1: Nicely done!

tests/slack_bolt/request/test_internals.py
```diff
@@ -286,6 +302,14 @@ def test_enterprise_id_extraction(self):
...
             enterprise_id = extract_enterprise_id(req)
             assert enterprise_id is None
+        for req in self.function_event_requests:
+            enterprise_id = extract_enterprise_id(req)
+            assert enterprise_id == "E111"
+
+    def test_function_execution_id_extraction(self):
+        for req in self.function_event_requests:
+            enterprise_id = extract_function_execution_id(req)
+            assert enterprise_id == "Fx111"
```
seratch, line 312:
```suggestion
            function_execution_id = extract_function_execution_id(req)
            assert function_execution_id == "Fx111"
```

seratch (changes requested):
Thanks for sharing this! This is already amazing but I left a few comments

slack_bolt/request/internals.py
WilliamBergamin: Good point 💯  I've implemented this in the `build_context` and `build_async_context` functions to reduce unnecessary runtime execution
The `bot_access_token` will only be parsed if there it a `function_execution_id` found in the payload

slack_bolt/context/base_context.py
```diff
@@ -103,13 +105,23 @@ def matches(self) -> Optional[Tuple]:
         """Returns all the matched parts in message listener's regexp"""
         return self.get("matches")
 
+    @property
+    def function_execution_id(self) -> Optional[str]:
+        """The `function_execution_id` associated with this request."""
```
filmaj, line 110: Should we document that this is only available for `function_executed` events and interactivity events scoped to a function?

slack_bolt/context/base_context.py
```diff
@@ -103,13 +105,23 @@ def matches(self) -> Optional[Tuple]:
...
     # --------------------------------
 
     @property
     def authorize_result(self) -> Optional[AuthorizeResult]:
         """The authorize result resolved for this request."""
         return self.get("authorize_result")
 
+    @property
+    def function_bot_access_token(self) -> Optional[str]:
+        """The bot token resolved for this function request."""
```
filmaj, line 122: Same as above re: only available for `function_executed` events and interactivity events scoped to a function.

slack_bolt/listener_matcher/builtins.py
```diff
@@ -176,6 +177,17 @@ def _verify_message_event_type(event_type: str) -> None:
         raise ValueError(error_message_event_type(event_type))
 
 
+def function_event(
```
filmaj, line 180: Can we add a docstring here to clarify what this is for? It is clear to me given the context, but I don't trust my memory to remember this in e.g. a year from now 😅 

The event and construct here (Slack Custom Function) is also named in a way that makes it confusing when discussing it in a programming concept (are you talking about the Slack Custom Function, or a Python function? etc.). I think we need to be disciplined about terminology.

filmaj: Technically the `type` of the specific event related here is called a `function_executed` event - perhaps tweaking the naming to match that makes sense too.

misscoded (commented):
Nice work! Looking good so far. 🎉 

**Usability, is this what we envisioned the developer experience would be?**
Usability in the gist looks good, though I want to confirm that we're going to also support the same utilities currently available in the BoltJS implementation, correct? i.e., `inputs`, `definition`, `complete`, `fail`.

**Does this aline with the Bolt-js experience?**
So long as the same utilities are pulled over (see above) and we replace the low-level API calls (with formal methods (e.g., `functions_completeSuccess`), the experience looks generally aligned.

**Implementation, does this implementation align with bolt principles**
Enough, IMO, though Kaz to ultimately decide. 

There are some differences to be expected between the two, but that’s OK. For example, the `matchers` aspect is nested within the `WorkflowFunction` class in Bolt, but what you have here appears to be the standard approach in Bolt for Python. 

One thing: we have gone with a class-based design for tidiness/organization purposes (similar to `WorkflowStep` seen [here](https://github.com/slackapi/bolt-python/blob/676035a1862920ec72c5755df518a2797b79f65a/slack_bolt/app/app.py#L690C33-L690C33)). I ultimately leave it to Kaz to decide if he wants to carry this approach over.

**Tests, is this missing tests that could be valuable for the feature**
We currently have tests as a non-blocking requirement. That said, given how quickly we’re progressing thus far, I don’t see an issue with us ensuring tests are also introduced by EOQ.

filmaj (approved):
Hell yeah I love it!
The API looks great from my point of view - feels extremely familiar.
Worked like a charm too 🥇 
This is really exciting!!!

WilliamBergamin:
FYI, PR #987 should fix the failing tests

WilliamBergamin:
@misscoded thanks for the feedback 🙇 

>  I want to confirm that we're going to also support the same utilities currently available in the BoltJS implementation, correct? i.e., `inputs`, `definition`, `complete`, `fail`

We will definitely  support `inputs`, `complete` and `fail` 💯 as for  `definition` I need to confirm this is something we can extract from the payload, I think we can add these helpers in follow up PRs this should make it easier to review and priorities.  
This PR aims to be the minimal implementation of the function handler

slack_bolt/context/base_context.py
WilliamBergamin: Good point I will add something like "only available for function related events"?

slack_bolt/listener_matcher/builtins.py
WilliamBergamin: I like renaming this function to match the event `type` 💯  I agree that it aligns better with the terminology

seratch (approved):
looks great to me so far!

slack_bolt/logger/messages.py
```diff
@@ -266,7 +267,37 @@ def warning_unhandled_request(  # type: ignore
...
+        event = req.body.get("event", {})
+        event_type = event.get("type")
+        if is_function(req.body):
+            # @app.function
+            callback_id = event.get("function", {}).get("callback_id", "function_id")
+            return _build_unhandled_request_suggestion(
+                default_message,
+                f"""
+@app.function("{callback_id}")
+{'async ' if is_async else ''}def handle_{callback_id}_function(body, event, client, logger):
```
seratch, line 279: It seems that you can have "-" as part of callback_id. In this case, the generated code can have syntax error in Python. I haven't examined what else can be used for callback_id, so replacing all non-alphanumeric characters to "_" (underscore) should be a safe way to deal with such edge cases.

slack_bolt/middleware/__init__.py
```diff
@@ -37,5 +39,5 @@
     "RequestVerification",
     "SslCheck",
     "UrlVerification",
-    "builtin_middleware_classes",
+    "AttachingFunctionToken" "builtin_middleware_classes",
```
seratch, line 42:
```suggestion
    "AttachingFunctionToken",
    "builtin_middleware_classes",
```

seratch (commented):
while reviewing this PR again, i found a few minor issues

slack_bolt/logger/messages.py
WilliamBergamin: Good catch 💯 I've opted to follow a similar pattern found in this file, by hardcoding the function name as `handle_some_function` instead of using the callback_id to dynamically name the function

WilliamBergamin:
Changed the base branch from `main` to `feat-functions` in order to properly create a release candidate with the changes

PR #1012, opened 2024-01-12
Replace Flask-Sockets with aiohttp for testing 

WilliamBergamin:
This PR bring the logic from [slack-sdk #1445](https://github.com/slackapi/python-slack-sdk/pull/1445) to bolt python

`Flask-Sockets` [is deprecated](https://github.com/heroku-python/flask-sockets/issues/85) and it won't receive any fixes in the future, so it seems reasonable to replace it with some up-to-date library.
`aiohttp` has been chosen since it was already presented as an optional requirement.

* [x] Others

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -23,7 +23,7 @@ jobs:
         run: |
           pip install -U pip
           pip install -r requirements.txt
-          pip install -r requirements/testing_without_asyncio
+          pip install -r requirements/testing_without_asyncio.txt
       - name: Run tests without aiohttp
         run: |
           pytest tests/slack_bolt/
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -23,7 +23,7 @@ jobs:
         run: |
           pip install -U pip
           pip install -r requirements.txt
-          pip install -r requirements/testing.txt
+          pip install -r requirements/testing_without_asyncio
       - name: Run tests without aiohttp
         run: |
           pytest tests/slack_bolt/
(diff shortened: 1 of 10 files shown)


.github/workflows/tests.yml
```diff
@@ -23,7 +23,7 @@ jobs:
         run: |
           pip install -U pip
           pip install -r requirements.txt
-          pip install -r requirements/testing_without_asyncio.txt
```
seratch, line 26: Can you revert this change? This part intentionally does not have aiohttp to detect unexectedly having aiohttp depedency to the core part  (as the following name "Run tests without aiohttp" indicate)

seratch (changes requested):

.github/workflows/tests.yml
WilliamBergamin: I've reverted this change but `aoihttp` will be required to run `pytest tests/adapter_tests/socket_mode/`, we can try using another synchronous library or move the CI step `Run tests for Socket Mode adapters` after `Run tests for HTTP Mode adapters (asyncio-based libraries)` let me know what you think

seratch: > pytest tests/adapter_tests/socket_mode/

Sorry, I should have clearly mentioned this but adding aiohttp for these tests is totally. My concern was about tests under slack_sdk etc.

WilliamBergamin: Ohh I see, I've update the ci tests and dependencies to make them pass, let me know if we prefer a sync approach to this?

seratch: Thank you! The only risk is that we may unintentionally add aiohttp imports to the non-asyncio Socket Mode adapters. Thus, making the changes only to the adapter tests should be safe enough. We can detect such in code reviews.

seratch (approved):

PR #1017, opened 2024-01-19
feat: improve test speed

WilliamBergamin:
Flow up to the discussion started in #1015 

This PR aims to improvement on the `mock_web_api_server` by making it more robust and attempts to reduce the time spent sleeping in test ("Run tests without aiohttp") 

This is achieved by incorporating a solution that allows for waiting up to a specified duration (X amount of time) to confirm the receipt of requests, instead of waiting for a fixed period and then confirming the reception of requests.

Previously, the test located in "Run tests without aiohttp" required 93 seconds for execution. With these changes, the tests can now be completed in just 24 seconds, resulting in a nearly fourfold reduction in wait time.

Similar changes can be applied to asynchronous test and similar reduction in wait time can also be expected for this as well

NOTE: adding content length header to the `mock_web_api_server` drastically improved request reliability

#### Before
Screenshot 2024-02-07 at 11 17 07 AM

#### After
Screenshot 2024-02-07 at 11 17 28 AM

### Feedback
The code can be made clearer and there may be some changes that are not necessary, please point out anything that seems relevant

**Is this approach worth implementing?**
 
* [x] Others

### DIFF
--- a/tests/adapter_tests_async/test_async_asgi.py
+++ b/tests/adapter_tests_async/test_async_asgi.py
@@ -221,4 +221,4 @@ class TestAsyncAsgi:
 
         assert response.status_code == 200
         assert response.headers.get("content-type") == "application/json;charset=utf-8"
-        assert_auth_test_count(self, 1)
+        assert_auth_test_count(self, 0)
--- a/tests/mock_web_api_server.py
+++ b/tests/mock_web_api_server.py
@@ -237,7 +237,7 @@ class MockServerThread(threading.Thread):
         self.join()
 
 
-class ReceivedRequestsHandler:
+class ReceivedRequests:
     def __init__(self, queue: Queue):
         self.queue = queue
         self.received_requests = {}
@@ -251,8 +251,8 @@ class ReceivedRequestsHandler:
 
 def setup_mock_web_api_server(test: TestCase):
     test.server_started = threading.Event()
-    test.received_requests_handler = ReceivedRequestsHandler(Queue())
-    test.thread = MockServerThread(test.received_requests_handler.queue, test)
+    test.received_requests = ReceivedRequests(Queue())
+    test.thread = MockServerThread(test.received_requests.queue, test)
     test.thread.start()
     test.server_started.wait()
 
@@ -267,11 +267,11 @@ def assert_received_request_count(test: TestCase, path: str, min_count: int, tim
     error = None
     while time.time() - start_time < timeout:
         try:
-            assert test.received_requests_handler.get(path, 0) == min_count
+            assert test.received_requests.get(path, 0) == min_count
             return
         except Exception as e:
             error = e
-            # waiting for mock_received_requests updates
+            # waiting for some requests to be received
             time.sleep(0.05)
 
     if error is not None:
--- a/tests/mock_web_api_server.py
+++ b/tests/mock_web_api_server.py
@@ -302,21 +302,19 @@ def assert_auth_test_count(test: TestCase, expected_count: int):
     assert_received_request_count(test, "/auth.test", expected_count, 0.5)
 
 
-async def assert_received_request_count_async(test: TestCase, path: str, expected_count: int, timeout: float):
-    start_time = time.time()
+async def assert_auth_test_count_async(test: TestCase, expected_count: int):
+    await asyncio.sleep(0.1)
+    retry_count = 0
     error = None
-    while time.time() - start_time < timeout:
+    while retry_count < 3:
         try:
-            assert test.received_requests_handler.get(path, 0) == expected_count
-            return
+            test.mock_received_requests.get("/auth.test", 0) == expected_count
+            break
         except Exception as e:
             error = e
+            retry_count += 1
             # waiting for mock_received_requests updates
-            await asyncio.sleep(0.05)
+            await asyncio.sleep(0.1)
 
     if error is not None:
         raise error
-
-
-async def assert_auth_test_count_async(test: TestCase, expected_count: int):
-    await assert_received_request_count(test, "/auth.test", expected_count, 0.5)
--- a/tests/mock_web_api_server.py
+++ b/tests/mock_web_api_server.py
@@ -1,7 +1,6 @@
 import asyncio
 import json
 import logging
-import sys
 import threading
 import time
 from http import HTTPStatus
@@ -9,11 +8,6 @@ from http.server import HTTPServer, SimpleHTTPRequestHandler
 from typing import Type, Optional
 from unittest import TestCase
 from urllib.parse import urlparse, parse_qs, ParseResult
-
-from multiprocessing import Process
-from urllib.request import urlopen, Request
-
-from tests.utils import get_mock_server_mode
 
 
 class MockHandler(SimpleHTTPRequestHandler):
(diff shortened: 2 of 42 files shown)


tests/adapter_tests_async/test_async_asgi.py
```diff
@@ -221,4 +221,4 @@ async def test_url_verification(self):
 
         assert response.status_code == 200
         assert response.headers.get("content-type") == "application/json;charset=utf-8"
-        assert_auth_test_count(self, 1)
+        assert_auth_test_count(self, 0)
```
seratch, line 224: Indeed, this should be zero, but perhaps this had been wrong due to the current test runner mechanism's behavior, right? Nice improvement 👍

tests/mock_web_api_server.py
```diff
@@ -22,9 +23,9 @@ def is_valid_token(self):
     def is_valid_user_token(self):
         return "Authorization" in self.headers and str(self.headers["Authorization"]).startswith("Bearer xoxp-")
 
-    def set_common_headers(self):
+    def set_common_headers(self, content_length: int = 0):
         self.send_header("content-type", "application/json;charset=utf-8")
-        self.send_header("connection", "close")
+        self.send_header("Content-Length", str(content_length))
```
seratch, line 28: nit: As the above "content-type" works well, consistently using "content-length" (all lower case) would look sightly nicer to me
```suggestion
        self.send_header("content-length", str(content_length))
```

tests/mock_web_api_server.py
```diff
@@ -121,18 +122,19 @@ def set_common_headers(self):
...
         try:
             if path == "/webhook":
                 self.send_response(200)
-                self.set_common_headers()
+                self.set_common_headers(len("OK"))
                 self.wfile.write("OK".encode("utf-8"))
                 return
 
-            if path == "/received_requests.json":
+            if path == "/health":
```
seratch, line 134: Are you using this endpoint? If we no longer need this, you can simply remove it.

tests/mock_web_api_server.py
```diff
@@ -253,9 +257,22 @@ def stop(self):
         self.join()
 
 
+class ReceivedRequestsHandler:
+    def __init__(self, queue: Queue):
+        self.queue = queue
+        self.received_requests = {}
+
+    def get(self, key, default=None):
+        while not self.queue.empty():
```
seratch, line 266: This is quite a wise approach 👍

tests/mock_web_api_server.py
```diff
@@ -253,9 +257,22 @@ def stop(self):
         self.join()
 
 
+class ReceivedRequestsHandler:
+    def __init__(self, queue: Queue):
+        self.queue = queue
+        self.received_requests = {}
+
+    def get(self, key, default=None):
```
seratch, line 265:
```suggestion
    def get(self, key: str, default: Optional[int] =None) -> Optional[int]:
```

tests/mock_web_api_server.py
```diff
@@ -253,9 +257,22 @@ def stop(self):
         self.join()
 
 
+class ReceivedRequestsHandler:
```
seratch, line 260: this is a matter of preference but this class does not handle anything but it just stores the received requests. If I named this, it would be a simpler one such as "ReceivedRequests" or "SavedReceivedRequests"

tests/mock_web_api_server.py
```diff
@@ -253,9 +257,22 @@ def stop(self):
...
+        while not self.queue.empty():
+            path = self.queue.get()
+            self.received_requests[path] = self.received_requests.get(path, 0) + 1
+        return self.received_requests.get(key, default)
+
+
 def setup_mock_web_api_server(test: TestCase):
     test.server_started = threading.Event()
-    test.thread = MockServerThread(test)
+    test.received_requests_handler = ReceivedRequestsHandler(Queue())
```
seratch, line 274: For the same reason I mentioned above, `test.received_requests.get("/auth.test")` sounds more natural to me.

tests/scenario_tests/test_app.py
```diff
@@ -77,7 +74,7 @@ class TestExecutor(Executor):
 
     def test_valid_single_auth(self):
         app = App(signing_secret="valid", client=self.web_client)
-        assert app != None
+        assert app is not None
```
seratch, line 77: good catch

tests/scenario_tests/test_events_ignore_self.py
```diff
@@ -38,9 +39,9 @@ def handle_app_mention(say):
         response = app.dispatch(request)
         assert response.status == 200
         assert_auth_test_count(self, 1)
-        sleep(1)  # wait a bit after auto ack()
+        sleep(0.5)  # wait a bit after auto ack()
```
seratch, line 42: Is this still necessary?

seratch (commented):
THIS IS FANTASTIC! Thank you so much for drastically improving the speed and the stability too! We definitely should go with this approach!

tests/adapter_tests_async/test_async_asgi.py
WilliamBergamin: Yess, I suspect this may be caused by the current test runner mechanism's behavior but I need to dig into each individual test to see what the expected behavior should be and why these values are different, there seem to be a few like these 🤔

tests/mock_web_api_server.py
WilliamBergamin: Good point 💯  I did add this to try and see if it would be more reliable to ensure the `mock_web_api_server` responds with a success response before running the tests, there was no notable difference here

tests/scenario_tests/test_app.py
WilliamBergamin: True, but I think I'll remove these changes in order to simplify the PR

tests/scenario_tests/test_events_ignore_self.py
WilliamBergamin: Yess this is still required since we are asserting that an endpoint was not executed
If we validate without waiting we may assert the request was not sent before it gets sent

tests/mock_web_api_server.py
WilliamBergamin: Agree 💯

tests/mock_web_api_server.py
WilliamBergamin: Agree 💯

seratch (approved):
Looks great to me 👍 You can merge this PR at any time

WilliamBergamin:
@seratch I understood why some `assert_auth_test_count` need to be asserted as being called twice instead of once

1. The `auth test` is performed for the bot token and the user token in [InstallationStoreAuthorize](https://github.com/slackapi/bolt-python/blob/7af76f99bf15953192276a80d9f783041ebf8e95/slack_bolt/authorization/authorize.py#L333-L336)
2. Many tests use a variation of [this](https://github.com/slackapi/bolt-python/blob/7af76f99bf15953192276a80d9f783041ebf8e95/tests/scenario_tests/test_app_bot_only.py#L52-L73) fake MemoryInstallationStore implementation, where a `user_token` is hard coded into the `installation` returned by `find_installation`

I'm assuming the changes regarding the auth_test assertion count now reflect the behavior of bolt python, but wondering if their may be an underlying issue, does the bot token need to be auth tested for each request or should this only be done once?

Issue #31, opened 2020-08-05
Add global error handler


seratch:
## Description

As with other Bolt frameworks, the error handler should be customizable.

https://github.com/slackapi/bolt-python/blob/3eea0c456e0f675d0cf3b1e3c8446f6ff140f397/slack_bolt/app/app.py#L321-L324

https://github.com/slackapi/bolt-python/blob/3eea0c456e0f675d0cf3b1e3c8446f6ff140f397/slack_bolt/app/async_app.py#L352-L355

### Category (place an `x` in each of the `[ ]`)

* [x] `slack_bolt.App` and/or its core components
* [x] `slack_bolt.async_app.AsyncApp` and/or its core components
* [ ] Adapters in `slack_bolt.adapter`

## Requirements (place an `x` in each of the `[ ]`)

* [x I've read and understood the [Contributing guidelines](https://github.com/slackapi/bolt-python/blob/main/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).
* [x] I've searched for any related issues and avoided creating a duplicate issue [here](https://github.com/slackapi/bolt-python/issues).

PR #1143, opened 2024-08-30
disc: remove the optional typing of client

WilliamBergamin:
This PR opens the discussion to why the `context.client` property is `Optional`

Looking back to #31 its seems like the [original implementation](https://github.com/slackapi/bolt-python/pull/43/files#diff-b0be37fba6306a268dcad5ef5c39c6b3404e0f66d759cc16cb4419a62b18023bR13-R16) of the client property could return `None`, but this implementation has now changed. 
Original implementation:
```
    @property
    def client(self) -> Optional[WebClient]:
        return self.get("client", None)
```

Can we now remove this `Optional` type annotation, I cannot come up with a use case where `context.client` could be `None`, let me know if missing context around this 🤔 

* [ ] Others

### DIFF
--- a/slack_bolt/context/async_context.py
+++ b/slack_bolt/context/async_context.py
@@ -41,7 +41,7 @@ class AsyncBoltContext(BaseContext):
         return self["listener_runner"]
 
     @property
-    def client(self) -> Optional[AsyncWebClient]:
+    def client(self) -> AsyncWebClient:
         """The `AsyncWebClient` instance available for this request.
 
             @app.event("app_mention")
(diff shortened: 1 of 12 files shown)


slack_bolt/context/async_context.py
```diff
@@ -41,7 +41,7 @@ def listener_runner(self) -> "AsyncioListenerRunner":  # type: ignore[name-defin
         return self["listener_runner"]
 
     @property
-    def client(self) -> Optional[AsyncWebClient]:
```
WilliamBergamin, line 44: Relevant change ⬆️

WilliamBergamin, line 44: Relevant change ⬆️

slack_bolt/context/context.py
```diff
@@ -42,7 +42,7 @@ def listener_runner(self) -> "ThreadListenerRunner":  # type: ignore[name-define
         return self["listener_runner"]
 
     @property
-    def client(self) -> Optional[WebClient]:
```
WilliamBergamin, line 45: Relevant change ⬆️

seratch (approved):
Thanks, indeed the the type can be non-optional now; looks good to me

PR #1254, opened 2025-02-28
Remove mention of assistants and replace with AI apps

haleychaas:
Updating mention of assistants and agents to be AI apps instead (agents are reserved for Agentforce).


lukegalbraithrussell (approved):
I'm not sure who this "helmendorf" is from your branch name, haleychaas

PR #1437, opened 2026-02-11
Add 'agent: BoltAgent' listener argument

mwbrooks:
## Changelog

_Do not include in our Changelog because this is marked as an experimental feature._

## Summary

This pull request adds a `agent: BoltAgent` listener argument, which will be used to access AI agent-related features such as `chat_stream` and more.

- Introduces `BoltAgent` and `AsyncBoltAgent` classes that provide a convenient `chat_stream()` method pre-configured with event context defaults (`channel_id`, `thread_ts`, `team_id`, `user_id`)
- Wires agent into Bolt's `kwargs` injection system so listeners can declare it as a parameter (e.g. `def handle(agent: BoltAgent)`) or access it via `context.agent`
- `BoltAgent` construction is deferred - only created when the listener actually requests the agent or args parameter, avoiding unnecessary object creation on every dispatch

## Experimental

The `agent: BoltAgent` class and listener argument are marked as an experimental feature. We are categorizing experimental features as `semver:patch` until the experimental status is removed.

No exists tests were modified and the goal is to be a low risk feature to add to `main`.

## Example

```python
@app.event("app_mention")
def handle_mention(agent: BoltAgent):
    stream = agent.chat_stream()
    stream.append(markdown_text="Hello!")
    stream.stop()
```

## Known limitations

`chat_stream()` currently only works when `thread_ts` is available in the event context (DMs and threaded replies). Top-level channel messages do not have `thread_ts`, and `ts` is not yet provided to `BoltAgent` - tracked as a `FIXME` in the code. I've experimented with a solution, but I'd prefer to introduce it in a separate PR so that we can review it independently.

### Testing

[→ Modified Sample App with `agent: BoltAgent` and `agent.chat_stream()`](https://github.com/slack-samples/bolt-python-assistant-template/compare/main...feat-agent-argument?expand=1)

```bash
# Create a modified copy of AI Agent Sample app that uses `agent.chat_stream()`
$ slack create my-app --template https://github.com/slack-samples/bolt-python-assistant-template --branch feat-agent-argument
$ cd my-app/

# Setup the app
$ python3 -m venv .venv
$ source .venv/bin/activate
$ pip install -r requirements.txt

# Run the app
$ slack run

# Test the app
# → Open DM → Send a message → Confirm it works
# → @Mention Reply in thread → Confirm it works

# Clean up
$ slack delete -f
$ cd ..
$ rm -rf my-app/
```

### Category

* [ ] Others

## Requirements 

### DIFF
--- a/slack_bolt/agent/agent.py
+++ b/slack_bolt/agent/agent.py
@@ -63,9 +63,10 @@ class BoltAgent:
             raise ValueError(
                 "Either provide all of channel, thread_ts, recipient_team_id, and recipient_user_id, or none of them"
             )
+        # Argument validation is delegated to chat_stream() and the API
         return self._client.chat_stream(
-            channel=channel or self._channel_id,
-            thread_ts=thread_ts or self._thread_ts,
+            channel=channel or self._channel_id,  # type: ignore[arg-type]
+            thread_ts=thread_ts or self._thread_ts,  # type: ignore[arg-type]
             recipient_team_id=recipient_team_id or self._team_id,
             recipient_user_id=recipient_user_id or self._user_id,
             **kwargs,
--- a/slack_bolt/agent/async_agent.py
+++ b/slack_bolt/agent/async_agent.py
@@ -60,9 +60,10 @@ class AsyncBoltAgent:
             raise ValueError(
                 "Either provide all of channel, thread_ts, recipient_team_id, and recipient_user_id, or none of them"
             )
+        # Argument validation is delegated to chat_stream() and the API
         return await self._client.chat_stream(
-            channel=channel or self._channel_id,
-            thread_ts=thread_ts or self._thread_ts,
+            channel=channel or self._channel_id,  # type: ignore[arg-type]
+            thread_ts=thread_ts or self._thread_ts,  # type: ignore[arg-type]
             recipient_team_id=recipient_team_id or self._team_id,
             recipient_user_id=recipient_user_id or self._user_id,
             **kwargs,


slack_bolt/kwargs_injection/async_utils.py
```diff
@@ -29,7 +29,7 @@ def build_async_required_kwargs(
     error: Optional[Exception] = None,  # for error handlers
     next_keys_required: bool = True,  # False for listeners / middleware / error handlers
 ) -> Dict[str, Any]:
-    all_available_args = {
+    all_available_args: Dict[str, Any] = {
```
mwbrooks: note: This fixed a linter warning

slack_bolt/kwargs_injection/async_utils.py
```diff
@@ -102,7 +106,7 @@ def build_async_required_kwargs(
     for name in required_arg_names:
         if name == "args":
             if isinstance(request, AsyncBoltRequest):
-                kwargs[name] = AsyncArgs(**all_available_args)  # type: ignore[arg-type]
+                kwargs[name] = AsyncArgs(**all_available_args)
```
mwbrooks: note: The above fix allows us to remove this type ignore

mwbrooks (commented):
Comments for the kind and knowledgeable reviewers!

slack_bolt/__init__.py
```diff
@@ -21,6 +21,7 @@
 from .response import BoltResponse
 
 # AI Agents & Assistants
+from .agent import BoltAgent
```
WilliamBergamin: 🤔 Maybe we should wait until the feature is GA before exporting it here?

Developers should still be able to import the class directly with something like
`from slack_bolt.agent import BoltAgent`

tests/scenario_tests/test_events_agent.py
```diff
@@ -0,0 +1,247 @@
...
+            recipient_user_id="U999",
+        )
+
+        client.chat_stream.assert_called_once_with(
+            channel="C999",
+            thread_ts="9999999999.999999",
+            recipient_team_id="T999",
+            recipient_user_id="U999",
+        )
+        assert stream is not None
```
WilliamBergamin: What do you think about creating a `tests.slack_bolt.agent` and `tests.slack_bolt_async.agent` directory where unit tests dedicated to the `BoltAgent` like these ones can live?

slack_bolt/kwargs_injection/async_utils.py
```diff
@@ -83,6 +83,10 @@ def build_async_required_kwargs(
         if k not in all_available_args:
             all_available_args[k] = v
 
+    # Defer agent creation to avoid constructing AsyncBoltAgent on every request
+    if "agent" in required_arg_names or "args" in required_arg_names:
+        all_available_args["agent"] = request.context.agent
```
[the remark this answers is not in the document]
mwbrooks: Great idea, @WilliamBergamin! I didn't know about `warnings` but it's a very nice way to let developers know that they're accessing an experimental feature.

Commit cf5ef985d890ecf1fd850328256bbdfc5d5cc98e adds the above suggestion and implements it. I added you as a co-author because I took your code verbatim. 👌🏻 

> IIUC every time a handler processes a request and uses the "agent" kwargs, this warning would be printed, this might be a bit annoying but it would be clear

It looks like there is a [Warning Filter](https://docs.python.org/3.7/library/warnings.html#the-warnings-filter) and the default is to only print the first occurrence. Personally, I'd be okay with printing on every occurrence since it's an experimental feature, but my implementation is using the default.

Here is a preview of the warning:
image

slack_bolt/__init__.py
mwbrooks: 🤔 If possible, I'd prefer to export the `BoltAgent` now.

1. Our work will be driven through developing a sample app, it would allow us to immediately understand what the production code will looks & feels like. [For example, here is what the import would look like today.](https://github.com/slack-samples/bolt-python-assistant-template/compare/main...feat-agent-argument?expand=1#diff-3f9dc6b27d130a458c6eeec2dfba277c7b6832098fa0a9fae0129badc5be5be1R5).
2. Our new `ExperimentalWarning` makes it clear that the developer is using something that's unstable and at their own risk.
3. When it comes to remove the Experimental status, I'd prefer to not have to add additional code such as a public export. I'd feel more confident about removing the Experimental status if we had tested everything throughout the entire process.

👉🏻 That said, I trust you decision over ours since you're more experienced with Bolt Python. If it makes you feel more comfortable waiting until the feature is GA, just let me know.

slack_bolt/context/async_context.py
```diff
@@ -15,6 +15,9 @@
 from slack_bolt.context.set_title.async_set_title import AsyncSetTitle
 from slack_bolt.util.utils import create_copy
 
+if TYPE_CHECKING:
+    from slack_bolt.agent.async_agent import AsyncBoltAgent
+
```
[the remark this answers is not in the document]
mwbrooks: > Static type checking around AsyncBoltAgent passes

This was my main motivation to add it. At first, I felt like it was the wrong approach, but it seems be correct and necessary with the Async modules.

slack_bolt/agent/agent.py
```diff
@@ -0,0 +1,82 @@
...
+        resolved_channel = channel or self._channel_id
+        resolved_thread_ts = thread_ts or self._thread_ts
+        if resolved_channel is None:
+            raise ValueError(
+                "channel is required: provide it as an argument or ensure channel_id is set in the event context"
+            )
+        if resolved_thread_ts is None:
+            raise ValueError(
+                "thread_ts is required: provide it as an argument or ensure thread_ts is set in the event context"
+            )
```
zimeg:
```suggestion
```

🪓 thought(non-blocking): Perhaps against above comments too, but I have small preference to pass missing values to the API for errors from upstream. I realize though that catching error earlier might be preferred, and we have validation logic for similar parameters too!

tests/scenario_tests/test_events_agent.py
```diff
@@ -0,0 +1,189 @@
...
+            assert agent is not None
+            assert isinstance(agent, BoltAgentDirect)
+            # Verify the same instance is returned on subsequent access
+            assert context.agent is agent
+            state["called"] = True
+
+        request = BoltRequest(body=app_mention_event_body, mode="socket_mode")
+        response = app.dispatch(request)
+        assert response.status == 200
+        assert_target_called()
```
zimeg: 🧪 note: Ahaha I notice we might be expecting `context` and `kwargs` arguments to match for the agent! Thanks for including these tests!

slack_bolt/context/async_context.py
```diff
@@ -187,6 +190,34 @@ async def handle_button_clicks(context):
...
+            async def handle_mention(agent):
+                stream = await agent.chat_stream()
+                await stream.append(markdown_text="Hello!")
+                await stream.stop()
+
+        Returns:
+            `AsyncBoltAgent` instance
+        """
+        if "agent" not in self:
+            from slack_bolt.agent.async_agent import AsyncBoltAgent
```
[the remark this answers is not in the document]
mwbrooks: Thanks for catching this @zimeg. I agree, I'd probably be the developer who "refactored" the code to readability only to introduce an unintended side effect. Commit c2a590d adds a comment.

I'm not 100% certain that we'll keep the `agent` argument. If we do, then we'll want to revisit @WilliamBergamin's comment about removing the dynamic import to avoid runtime errors.

slack_bolt/warning/__init__.py
```diff
@@ -0,0 +1,7 @@
+"""Bolt specific warning types."""
+
+
+class ExperimentalWarning(FutureWarning):
+    """Warning for features that are still in experimental phase."""
+
```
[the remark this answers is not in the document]
mwbrooks: 🫡 Roger! Commit 0195ee2 removes the whitespace. I'm still a little unclear on when Python wants/requires empty lines and when it doesn't.

PR #1481, opened 2026-04-11
fix: handle malformed user/view payloads in extract_team_id

### LINKED ISSUE #1447: `extract_team_id` crashes with `TypeError when payload["user"]` is a string (e.g. `member_joined_channel` events)
#### The `slack_bolt` version

```
slack-bolt==1.27.0
slack-sdk==3.39.0
```

#### Python runtime version

```
Python 3.14.0
```

#### OS info

```
ProductName:    macOS
ProductVersion: 26.3
BuildVersion:   25D125
Darwin Kernel Version 25.3.0: Wed Jan 28 20:53:05 PST 2026; root:xnu-12377.81.4~5/RELEASE_ARM64_T6020
```

#### Steps to reproduce:

1. Register a `member_joined_channel` listener in a Slack Bolt app:

```python
@app.event("member_joined_channel")
def handle_channel_join(event, client, say):
    say("Hello!")
```

2. Add the bot to any Slack channel.

3. Observe the incoming event payload — Slack sends `"user"` as a plain string (the user ID):

```json
{
  "type": "member_joined_channel",
  "user": "U0123456789",
  "channel": "C0123456789",
  "team": "T0123456789"
}
```

### Expected result:

The listener is invoked normally.

### Actual result:

The request crashes before the listener is reached, with:

```
TypeError: string indices must be integers, not 'str'

File "slack_bolt/adapter/aws_lambda/handler.py", line 87, in to_bolt_request
    return BoltRequest(
File "slack_bolt/request/request.py", line 69, in __init__
    self.context = build_context(BoltContext(context if context else {}), self.body)
File "slack_bolt/request/internals.py", line 276, in build_context
    team_id = extract_team_id(body)
File "slack_bolt/request/internals.py", line 116, in extract_team_id
    return payload["user"]["team_id"]
```

### Root cause

`extract_team_id` in `slack_bolt/request/internals.py` assumes `payload["user"]` is always a dict:

```python
if payload.get("user") is not None:
    return payload["user"]["team_id"]
```

But `member_joined_channel` sends `"user"` as a plain string (the user ID). Accessing `"U0123456789"["team_id"]` raises the TypeError.

The same function already handles `payload["team"]` correctly with an `isinstance` check:

```python
if payload.get("team") is not None:
    team = payload.get("team")
    if isinstance(team, str):
        return team
    elif team and "id" in team:
        return team.get("id")
```

The same guard is missing for `payload["user"]`. This pattern (subscribing to `member_joined_channel` and checking if the joined user is the bot) is what maintainers recommend in #710, so users following that guidance will hit this bug.

saschabuehrle:
## Summary
This prevents `extract_team_id()` from raising `KeyError` when malformed payloads contain `user` or `view` objects without a `team_id` key.

## Changes
- use `.get("team_id")` for `payload["user"]` and `payload["view"]` lookup in `extract_team_id`
- extend invalid-payload coverage in `test_extraction_functions_invalid_dict_keys`

## Why
Slack requests should still be rejected by request verification, but malformed traffic should not crash context extraction before middleware runs.

## Testing
- `python3 -m compileall -q slack_bolt/request/internals.py tests/slack_bolt/request/test_internals.py`

Fixes #1447

Greetings, saschabuehrle

### DIFF
--- a/slack_bolt/request/internals.py
+++ b/slack_bolt/request/internals.py
@@ -112,9 +112,9 @@ def extract_team_id(payload: Dict[str, Any]) -> Optional[str]:
     if isinstance(payload.get("event"), dict):
         return extract_team_id(payload["event"])
     if isinstance(payload.get("user"), dict):
-        return payload["user"]["team_id"]
+        return payload["user"].get("team_id")
     if isinstance(payload.get("view"), dict):
-        return payload["view"]["team_id"]
+        return payload["view"].get("team_id")
     return None
 
 
--- a/tests/slack_bolt/request/test_internals.py
+++ b/tests/slack_bolt/request/test_internals.py
@@ -1253,8 +1253,10 @@ class TestRequestInternals:
         invalid_payloads = {
             "event": {"event": "some_event_type"},
             "user": {"user": "U12345"},
+            "user_missing_team_id": {"user": {"id": "U12345"}},
             "team": {"team": "T12345"},
             "view": {"view": "V12345"},
+            "view_missing_team_id": {"view": {"id": "V12345"}},
             "message": {"message": "some text"},
             "item": {"item": "item_id"},
             "function_data": {"function_data": "fd_123"},


WilliamBergamin (approved):
Thanks for catching this edge case @saschabuehrle & contributing 🥇 change look good to me!

The optimal fix for #1447 is to get the `RequestVerification` middleware to run before the extraction logic, this would involve moving the extraction logic into a middleware which would be a breaking change

This PR in combination with #1464 aim to improve the robustness of the extraction logic so that no errors are thrown before the `RequestVerification` middleware is ran, for this reason we won't close #1447  just yet

PR #1495, opened 2026-05-08
fix: Require signatures for ssl_check request verification

### LINKED ISSUE #1494: fix: Request verification can be skipped with ssl_check=1 when SSL check middleware is disabled
When an app is configured with `request_verification_enabled=True` and `ssl_check_enabled=False`, a form-encoded request containing `ssl_check=1` skips Slack signature verification.

`RequestVerification._can_skip` currently treats any parsed body with `ssl_check=1` as safe to bypass, but that check is independent of whether the `SslCheck` middleware is enabled. If the SSL-check middleware is disabled, the request can continue through the normal middleware/listener pipeline without a valid Slack signature.

Expected behavior: `ssl_check=1` should only bypass normal dispatch when the SSL-check middleware handles the request. If SSL-check handling is disabled, request verification should still require a valid Slack signature.

This affects both sync and async request verification paths.

homanp:
## Summary
- stop `RequestVerification` from skipping HMAC validation solely because a form body contains `ssl_check=1`
- keep the existing Socket Mode verification skip
- add regression coverage for sync, async, and WSGI dispatch paths

## Why
When `ssl_check_enabled=False`, the SSL-check middleware does not intercept `ssl_check=1` requests. Request verification still needs to authenticate those HTTP requests before they can reach normal middleware or slash command listeners.

Fixes #1494.

## Testing
- `./scripts/run_tests.sh "tests/slack_bolt/middleware/request_verification/test_request_verification.py tests/slack_bolt_async/middleware/request_verification/test_request_verification.py tests/scenario_tests/test_slash_command.py tests/scenario_tests_async/test_slash_command.py tests/adapter_tests/wsgi/test_wsgi_http.py"`
- `./scripts/lint.sh --no-install`
- `./scripts/run_mypy.sh --no-install`

### DIFF
--- a/slack_bolt/middleware/request_verification/request_verification.py
+++ b/slack_bolt/middleware/request_verification/request_verification.py
@@ -49,7 +49,7 @@ class RequestVerification(Middleware):
 
     @staticmethod
     def _can_skip(mode: str, body: Dict[str, Any]) -> bool:
-        return mode == "socket_mode" or (body is not None and body.get("ssl_check") == "1")
+        return mode == "socket_mode"
 
     @staticmethod
     def _build_error_response() -> BoltResponse:
--- a/tests/adapter_tests/wsgi/test_wsgi_http.py
+++ b/tests/adapter_tests/wsgi/test_wsgi_http.py
@@ -89,6 +89,47 @@ class TestWsgiHttp:
         assert response.headers.get("content-type") == "text/plain;charset=utf-8"
         assert_auth_test_count(self, 1)
 
+    def test_ssl_check_param_does_not_bypass_request_verification(self):
+        app = App(
+            client=self.web_client,
+            signing_secret=self.signing_secret,
+            ssl_check_enabled=False,
+        )
+        command_called = False
+
+        def command_handler(ack):
+            nonlocal command_called
+            command_called = True
+            ack()
+
+        app.command("/hello-world")(command_handler)
+
+        body = (
+            "token=verification_token"
+            "&team_id=T111"
+            "&team_domain=test-domain"
+            "&channel_id=C111"
+            "&channel_name=random"
+            "&user_id=W111"
+            "&user_name=primary-owner"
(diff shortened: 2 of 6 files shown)


homanp:
CLA was signed but not showing up here.

homanp:
@WilliamBergamin found this as well, not sure if this was "by-design" or not but worth taking a look.

WilliamBergamin (approved):
This is another awesome find 💯 🚀 

Reproducing this locally and testing it out took me longer then I want to admit, but changes look good to me 🥇 thanks for adding these unit tests, they will prevent us form introducing similar behavior in the future

WilliamBergamin:
This might of been by design to allows users that disable `SslCheck` middleware to somehow handle these requests manually through a handler 🤔 this feel like a super edge case that leaves the door wide open 

User that want to handle their own ssl check logic are able to do so in a global middleware

Its also safe to assume that users that disable `SslCheck` middleware are dealing with advanced use cases and know what they are doing

homanp:
> This might of been by design to allows users that disable `SslCheck` middleware to somehow handle these requests manually through a handler 🤔 this feel like a super edge case that leaves the door wide open 
> 
> 
> 
> User that want to handle their own ssl check logic are able to do so in a global middleware
> 
> 
> 
> Its also safe to assume that users that disable `SslCheck` middleware are dealing with advanced use cases and know what they are doing

Hmm, do you have those use cases?

homanp:
> This is another awesome find 💯 🚀 
> 
> Reproducing this locally and testing it out took me longer then I want to admit, but changes look good to me 🥇 thanks for adding these unit tests, they will prevent us form introducing similar behavior in the future 

Happy to help

WilliamBergamin:
> Hmm, do you have those use cases?

I do not have these use cases, but there is probably someone out there with one, if these changes causes a problem for them, they are more then welcome to open an issue

PR #1517, opened 2026-05-31
Add async optional dependency extra

### LINKED ISSUE #1472: Fully define dependencies for async support via optional dependency group
The current docs suggest adding `aiohttp` as a dependency to create an async app, but this creates a direct dependency on `aiohttp` for consumers of `slack-bolt` on what is is really a transitive dependency for the consumer of `slack-bolt`. This is especially important because this leaves it to consumers to find compatible versions of `slack-bolt` and `aiohttp` without the aide of a package manager.

`slack-bolt` should define an optional dependency group and include `aiohttp` so that consumers of this library don't need to maintain a dependency on a transitive. Consumers would be able to `uv add slack-bolt[async]` for example to include the optional dependencies.

* [ ] **slack_bolt.App** and/or its core components
* [X] **slack_bolt.async_app.AsyncApp** and/or its core components
* [] Others

## Requirements

Cr1stal:
## Summary

Fixes #1472 by adding a public `async` optional dependency group for installing Bolt's async support:

- adds `requirements/async.txt` as the public dependency source for the extra
- exposes `slack_bolt[async]` through package metadata
- updates async installation docs and the `AsyncApp` module docstring

#### Reuse `requirements/async_dev.txt`

Rejected because `async_dev.txt` is a development/test requirements file. Exposing it through package extras would turn test-only dependency constraints into a public installation contract, which was the concern raised in #1423.

#### Support Python 3.7 and 3.8 through older `aiohttp` ranges

Rejected for this PR to keep the new public contract small. Bolt still supports Python 3.7+, but current `aiohttp` releases require Python 3.9+. Adding legacy ranges would make Bolt responsible for maintaining compatibility with older `aiohttp` versions.

#### Add `aiohttp` as a core dependency

Rejected because async support is optional and the core package should keep its current runtime dependency surface.

#### Keep documenting manual `aiohttp` installation only

Rejected because it leaves users responsible for discovering compatible async dependencies, which is the problem described in #1472.

### Testing

- [x] `pip wheel . --no-deps --no-build-isolation`
- [x] `./scripts/format.sh --no-install`
- [x] `./scripts/lint.sh --no-install`
- [x] `./scripts/run_tests.sh tests/slack_bolt_async`
- [x] `./scripts/run_mypy.sh --no-install`
- [x] `./scripts/install_all_and_run_tests.sh` (`910 passed, 81 warnings`)

### Category

* [x] Others

## Requirements

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,17 +150,18 @@ Most of the app's functionality will be inside listener functions (the `fn` para
 
 ## Creating an async app
 
-If you'd prefer to build your app with [asyncio](https://docs.python.org/3/library/asyncio.html), you can import the [AIOHTTP](https://docs.aiohttp.org/en/stable/) library and call the `AsyncApp` constructor. Within async apps, you can use the async/await pattern.
+If you'd prefer to build your app with [asyncio](https://docs.python.org/3/library/asyncio.html), you can install the async extra to include [AIOHTTP](https://docs.aiohttp.org/en/stable/) and call the `AsyncApp` constructor. Within async apps, you can use the async/await pattern.
 
 ```bash
-# Python 3.7+ required
+# Python 3.9+ required for the async extra
 python -m venv .venv
 source .venv/bin/activate
 
 pip install -U pip
-# aiohttp is required
-pip install slack_bolt aiohttp
+pip install "slack_bolt[async]"
 ```
+
+If you're using Python 3.7 or 3.8, install an `aiohttp` version compatible with your Python runtime separately.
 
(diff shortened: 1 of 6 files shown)


WilliamBergamin:
Hi @Cr1stal thanks again for your interest

Before we can implement these changes the team needs to figure out the support pattern for this, since all our dependencies (`slack_sdk`) support python `3.7+` I'm not sure its would be a common pattern to make our optional dependencies only support python `3.9+` 🤔

Issue #1346, opened 2025-08-05
Enhancement: Assistant should inherit global middleware for security and consistency


manthram:
## Problem
Assistant handlers bypass global middleware entirely, creating a significant security gap and architectural inconsistency. Assistant events, assistant.user_message, assistant.thread_started etc. skip critical middleware that regular events receive, including:

- SSL certificate verification
- Request signature verification
- Authorization middleware
- URL verification
- Custom middleware

This forces developers to manually implement security checks in each assistant handler, which is error-prone and violates the principle of secure-by-default design.

```
# These handlers get full middleware protection:
@app.event("app_mention")
async def handle_mention(event, say, context):
    # Already authenticated via middleware
    pass

# These handlers bypass ALL middleware (security vulnerability):

@assistant.user_message
async def handle_message(event, say, context):
    # No authentication, no SSL check, no signature verification!
    pass
```

Developer Experience: Developers must remember to add security manually to every assistant handler:

```
# Current: Manual security in every handler (error-prone)
@assistant.user_message(middleware=[auth_middleware, ssl_middleware])
async def handle_message(event, say, context):
    pass
```

### Category

* [X] **slack_bolt.App** and/or its core components
* [X] **slack_bolt.async_app.AsyncApp** and/or its core components
* [ ] Adapters in **slack_bolt.adapter**
* [ ] Others

### Requirements

Proposed Solution
Enhance Assistant/AsyncAssistant to automatically inherit the app's global middleware when handlers are registered. This would:

1. Maintain backwards compatibility - existing explicit middleware still works
2. Provide opt-in enhancement - controlled via auto_inherit_app_middleware=True parameter
3. Apply middleware in correct order - app middleware first, then handler-specific middleware

### Benefits

1. Security by default - Assistant events get same protection as regular events
2. Architectural consistency - All Slack events treated uniformly
3. Developer productivity - No more manual security boilerplate
4. Backwards compatible - Existing code continues working unchanged
5. Performance neutral - Middleware already exists, just applied consistently

PR #1519, opened 2026-05-31
Add explicit Assistant listener registration mode

Cr1stal:
## Summary

Addresses #1346 by adding an explicit Assistant registration mode to `App.assistant()` and `AsyncApp.assistant()`.

By default, `app.assistant(assistant)` continues to behave as it does today: it registers Assistant as middleware. This preserves the existing Assistant middleware dispatch path and keeps `app.use(assistant)` / `app.middleware(assistant)` behavior unchanged.

Apps can now opt into listener-based registration:

```python
app.assistant(assistant, mode="listeners")
```

In this mode, Assistant handlers are registered as normal App listeners. This lets assistant-specific handlers participate in the App listener pipeline and inherit app-level middleware such as authorization, auditing, rate limiting, or request enrichment.

The PR also adds an internal listener registry so Assistant listeners can be evaluated before broad catch-all listeners without exposing a general listener priority API. Registration order is preserved within Assistant listeners and ordinary App listeners, and sync/async behavior is kept mirrored.

### Testing

- `./scripts/format.sh --no-install`
- `./scripts/lint.sh --no-install`
- `./scripts/run_tests.sh tests/slack_bolt/app/test_app_assistant_middleware.py` (`14 passed`)
- `./scripts/run_mypy.sh --no-install` (`Success: no issues found in 235 source files`)
- `./scripts/install_all_and_run_tests.sh` (`924 passed`, `81 warnings`; mypy: `Success: no issues found in 235 source files`)

### Category 

* [ ] Others

## Requirements 

### DIFF
--- a/slack_bolt/app/app.py
+++ b/slack_bolt/app/app.py
@@ -681,10 +681,6 @@ class App:
             middleware_or_callable = args[0]
             if isinstance(middleware_or_callable, Middleware):
                 middleware: Middleware = middleware_or_callable
-                if isinstance(middleware, Assistant) and middleware.auto_inherit_app_middleware is True:
-                    # In this opt-in mode, Assistant handlers should run through the app listener pipeline.
-                    self._register_assistant_listeners(middleware)
-                    return None
                 self._middleware_list.append(middleware)
                 if isinstance(middleware, Assistant) and middleware.thread_context_store is not None:
                     self._assistant_thread_context_store = middleware.thread_context_store
@@ -713,8 +709,13 @@ class App:
     # -------------------------
     # AI Agents & Assistants
 
-    def assistant(self, assistant: Assistant) -> Optional[Callable]:
-        return self.middleware(assistant)
+    def assistant(self, assistant: Assistant, mode: str = "middleware") -> Optional[Callable]:
+        if mode == "middleware":
+            return self.middleware(assistant)
+        if mode == "listeners":
+            self._register_assistant_listeners(assistant)
+            return None
+        raise BoltError(f"Unsupported Assistant registration mode ({mode})")
 
     # -------------------------
     # Workflows: Steps from apps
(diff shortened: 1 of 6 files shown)


WilliamBergamin (commented):
Hi @Cr1stal thanks for your time working on this issue & interest in our project 💯 

Unfortunately I can't accept these changes, they introduce more complexity to the already complex `assistant` logic, this increases the maintenance burden on our team 😅 and can slow down future feature implementation

The ideal solution to #1346 might be to make the `assistant middleware` a none global middleware, but this might require some server side changes

All this to say we'd prefer to solve #1346 by simplifying the the `assistant` logic rather then increasing it's complexity

Cr1stal:
@WilliamBergamin Thanks, that makes sense. I understand the main concern is not the goal of #1346, but that this PR adds a second Assistant dispatch path and listener ordering machinery, which increases maintenance cost.

When you say the ideal solution might be making the Assistant middleware non-global, do you mean moving Assistant handling closer to the normal app listener pipeline while avoiding a new public registration mode and priority registry?

Also, could you clarify what server-side change you have in mind? My current read is that assistant user messages are hard to distinguish from broad message listeners without either server-side classification or client-side ordering rules.

WilliamBergamin:
> When you say the ideal solution might be making the Assistant middleware non-global, do you mean moving Assistant handling closer to the normal app listener pipeline while avoiding a new public registration mode and priority registry?

Yes, the goal would be to dissolve the assistant abstraction into the normal app listener pipeline. Users would configure their Bolt app as an 'assistant' and use the same app listener pipeline, rather than having a dedicated one for this abstraction.

> could you clarify what server-side change you have in mind? My current read is that assistant user messages are hard to distinguish from broad message listeners without either server-side classification or client-side ordering rules

Your current read is accurate. The biggest challenge in this space is around context handling; it's hard to distinguish assistant and user messages from broader messages. Server-side classification would be the ideal solution for this.

